PR #9, opened 2017-03-13
Fix build for CMake 2.8.12

pietern:
This is the minimum required CMake version (also the version that is available on Ubuntu Trusty (14.04)).


Yangqing:
Ah, with BUILD_TEST you will also need to install gtest right? I don't think this is currently done in the install script. Others look good

pietern:
@Yangqing Yeah, you're right. Changing to use a submodule for it. Quickly checked whether or not we could use the Ubuntu version but it's too old (1.6, we need 1.8).

pietern:
Rebase on 86ede33035478aa68f91182883ce80069e88333f

PR #10, opened 2017-03-14
Prevent tcp pair form throwing exceptions in the destructor

apaszke:
Previously when e.g. a Redis connection/store (e.g. protected mode) failed, it would throw an exception that would cause the pair to be destructed, and it would throw another one because it wasn't in the connected state yet. It's quite annoying because it shadows the original problem.

### DIFF
--- a/gloo/rendezvous/context.cc
+++ b/gloo/rendezvous/context.cc
@@ -33,8 +33,7 @@ void Context::connectFullMesh(
       continue;
     }
 
-    auto pair = dev->createPair();
-    pairs[i] = std::move(pair);
+    pairs[i] = dev->createPair();
 
     // Store address for pair for this rank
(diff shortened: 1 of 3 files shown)


pietern:
Thanks, @apaszke. I would also be fine with removing the assertion altogether, if it only complicates the internal API. For the reason you mention it's fine if a pair is closed when not yet connected. All that matters if that the state only moves forward, which is already asserted in `changeState`.

pietern:
Heads up @apaszke, @andrewwdye changed this to no longer assert on state when closing [in this commit](https://github.com/facebookincubator/gloo/commit/30a9cf7a46886ebdda4817b45090d6f83341445c#diff-33142785043eff3f4351c25ee9fa3d8eR543).

Closing this PR, since this is now fixed.

PR #17, opened 2017-03-21
MPI: Duplicate MPI_Comm and allreduce maxLength as MPI_ UNSIGNED_LONG.

gchanan:
Some small MPI-related changes:
1) Instead of making an object copy of the MPI_Comm, call MPI_Comm_dup;
because the (passed-in) communicator is used later via the call to
connectFullMesh this guarantees that the communicator will not have been
freed by user before connectFullMesh is called.

2) Allreduce for maxLength is done on an unsigned long type; use the
corresponding MPI type.

### DIFF
--- a/gloo/mpi/context.cc
+++ b/gloo/mpi/context.cc
@@ -18,26 +18,29 @@
 namespace gloo {
 namespace mpi {
 
-static int MPICommSize(MPI_Comm comm) {
+static int MPICommSize(const MPI_Comm &comm) {
   int comm_size;
   auto error = MPI_Comm_size(comm, &comm_size);
   GLOO_ENFORCE(error == MPI_SUCCESS, "MPI_Comm_size: ", error);
   return comm_size;
 }
 
-static int MPICommRank(MPI_Comm comm) {
+static int MPICommRank(const MPI_Comm &comm) {
   int comm_rank;
   auto error = MPI_Comm_rank(comm, &comm_rank);
   GLOO_ENFORCE(error == MPI_SUCCESS, "MPI_Comm_rank: ", error);
   return comm_rank;
 }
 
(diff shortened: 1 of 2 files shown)


gloo/mpi/context.cc
```diff
@@ -18,26 +18,29 @@
...
   auto error = MPI_Comm_rank(comm, &comm_rank);
   GLOO_ENFORCE(error == MPI_SUCCESS, "MPI_Comm_rank: ", error);
   return comm_rank;
 }
 
-Context::Context(MPI_Comm comm)
+Context::Context(const MPI_Comm &comm)
     : ::gloo::Context(MPICommRank(comm), MPICommSize(comm)),
-      comm_(comm) {
+      comm_() {
```
pietern: Can you remove it from the initializer list?

gloo/mpi/context.cc
```diff
@@ -18,26 +18,29 @@
 namespace gloo {
 namespace mpi {
 
-static int MPICommSize(MPI_Comm comm) {
+static int MPICommSize(const MPI_Comm &comm) {
```
pietern: Nit: we use `const T& foo` everywhere.

gloo/mpi/context.cc
pietern: Can you remove it from the initializer list?

gloo/mpi/context.h
```diff
@@ -19,7 +19,7 @@ namespace mpi {
 
 class Context : public ::gloo::Context {
  public:
-  explicit Context(MPI_Comm comm);
+  explicit Context(const MPI_Comm &comm);
```
pietern: Same here.

pietern:
Thanks @gchanan!

PR #18, opened 2017-03-22
Don't add options to CUDA_NVCC_FLAGS if already set

pietern:
This may be the case when the Gloo CMake files are sources from a
parent project that has already imported CMake CUDA support. If these
checks are not performed then CUDA_NVCC_FLAGS might contain
conflicting options.

Verified this works while working on Gloo for Caffe2.

### DIFF
--- a/cmake/Cuda.cmake
+++ b/cmake/Cuda.cmake
@@ -71,6 +71,36 @@ function(gloo_select_nvcc_arch_flags out_variable)
 endfunction()
 
 ################################################################################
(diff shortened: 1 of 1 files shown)


pietern:
@bwasti cc'ed you since you're a CMake guru and this is the first CMake function I ever wrote.

pietern:
Fixed the actual append.

Also had to add a set with PARENT_SCOPE to make the resulting list leave function scope...

PR #19, opened 2017-03-22
Make static/shared configurable and install optional

pietern:
This makes it possible to embed Gloo in a project without CMake
installing Gloo headers and/or libraries, or having a runtime
dependency (and statically link to it).

Also:
* Install benchmark tools
* Statically link to NCCL if the bundled version is used


pietern:
Verified the benchmark binaries are now self contained (no more runtime NCCL dependency).

Also verified Caffe2 can compile Gloo ops and doesn't have to install shared libraries to make it work.

PR #25, opened 2017-04-13
Expose missing headers

VirrageS:


pietern:
Thank you!

We should test for this from Travis... maybe include some sample downstream project that we build using CMake against an installed version of Gloo. That would at least spot this type of installation bugs.

PR #30, opened 2017-04-22
Install missing headers.

hgaiser:
This PR installs missing include headers.


pietern:
Thanks for your contributions!

These paths are not yet covered by CI... so fell through the cracks. Would be good to at some point test "make install" by having a few example programs that compile/link against Gloo from an installed location, so we know this doesn't regress.

Until then, do you mind signing the CLA? Then I can merge these fixes.

hgaiser:
Yeah I think the CLA is not an issue but I should let someone sign who is authorized on behalf of the company I work at. I will look into that tomorrow.

Testing these cases sounds like a solid plan. I came across this issue because gloo was used as a submodule in caffe2 but I thought it was more logical to have it as a system package. Is gloo most often used like that, as a submodule? Why is that? Just curious :)

pietern:
Thanks, re: the CLA.

It's currently a submodule since the project is still fairly new and it's more convenient to track the dependency in Caffe2 itself. Eventually we'll have to properly version gloo as well and be more diligent with potential API breakages but for now, with just being used from C2, it's been convenient tracking it directly.

It is also used from PyTorch, although that's still a work in progress.

Are you planning on using it directly, or only through Caffe2? Or not at all? ;)

hgaiser:
Ah I see, that makes more sense. A neat feature of Arch Linux is the Arch User Repository, which makes it really easy to set up a system package which tracks a git repository. Solves precisely situations like this. Unfortunately that won't be an option for Ubuntu and such ;)

https://aur.archlinux.org/packages/gloo-git/

I was checking out what caffe2 could do and trying out some tutorials. Ended up fixing dependencies like gloo being searched for system wide before searching submodules. In other words I'm planning to use gloo indirectly :)

https://github.com/caffe2/caffe2/pull/382

hgaiser:
Removed installing of `gloo/common.h` since you deleted it ;)

CLA should be signed now.

pietern:
Thanks @hgaiser!

PR #31, opened 2017-04-24
Only build on 64-bit systems

### LINKED ISSUE #28: Build error on ubuntu 14.04
Compilation error: https://github.com/caffe2/caffe2/issues/359

pietern:
While it is theoretically possible to make Gloo work on 32-bit systems, it's unlikely anybody would ever use it on 32-bit systems. This removes the expectation that it should work...

Fixes #28

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -2,6 +2,11 @@ cmake_minimum_required(VERSION 2.8.12 FATAL_ERROR)
 
 project(gloo CXX C)
 
+# Gloo assumes 64-bit and doesn't run builds/tests for anything else.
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -2,6 +2,11 @@ cmake_minimum_required(VERSION 2.8.12 FATAL_ERROR)
 
 project(gloo CXX C)
 
+# Gloo assumes 64-bit and doesn't run builds/tests anything else.
```
andrewwdye: nit: builds/tests *for* anything else, or build/test anything else

PR #45, opened 2017-06-07
Fix NCCL directory typo

slayton58:
Was `nccl_INCLUDE_DIRS` instead of `nccl_INCLUDE_DIR`


wesolwsk:
Thanks for the fix, Simon. Can you also change the place where nccl_INCLUDE_DIRS is set in Findnccl.cmake. Right now the build is failing.

slayton58:
I assume you meant in `cmake/External/nccl.cmake` -- Fixed so now everyone should be using `nccl_INCLUDE_DIR`, at least there are no more instances of `nccl_INCLUDE_DIRS` in the code.

PR #46, opened 2017-06-16
Add basic API support for NCCL 2.0

slayton58:
\cc @pietern 
Minimal changes to allow gloo to compile and run with NCCL 2.0


pietern:
Thanks, @slayton58!

PR #49, opened 2017-06-22
CUDA 9

slayton58:
Adds basic CUDA 9 support, including adding Volta arch, and making appropriate modifications for half precision datatype changes


gloo/types.h
```diff
@@ -56,17 +56,27 @@ struct __attribute__((__aligned__(2))) float16 {
     return x == res.x;
   }
 #ifdef __CUDA_ARCH__
-  float16(half h) {
+  float16(__half h) {
```
pietern: It's OK the constructor for `half` is gone now? (don't know where that type is defined TBH)

cmake/Cuda.cmake
```diff
@@ -1,15 +1,15 @@
 # Known NVIDIA GPU achitectures Gloo can be compiled for.
 # This list will be used for CUDA_ARCH_NAME = All option
-set(gloo_known_gpu_archs "20 21(20) 30 35 50 52 60 61")
-set(gloo_known_gpu_archs7 "20 21(20) 30 35 50 52")
+set(gloo_known_gpu_archs "30 35 50 52 60 61 70")
+set(gloo_known_gpu_archs7 "30 35 50 52")
```
pietern: Could you add a `gloo_known_gpu_archs8` so that we don't break the CUDA 8 builds. AFAICT this is what's tripping up the Travis build.

gloo/types.h
slayton58: I'm not 100% sure either :) I've always known the type as `__half`, and that's what is defined in cuda_fp16.h

cmake/Cuda.cmake
slayton58: Done and pushed

gloo/types.h
wesolwsk: I think half and __half are interchangeable. cuda_fp16.h has the following code:

```
#ifndef CUDA_NO_HALF
typedef __half half;
typedef __half2 half2;
#endif /*CUDA_NO_HALF*/
```

Should we stick to one for consistency? I vote for half. :)

slayton58: Yeah, some more directed grepping found that for me too. I don't care which one is used

pietern:
Thanks for adding this @slayton58!

PR #61, opened 2017-08-03
Multi-threading for benchmark tool

pietern:
This change makes it possible to run a particular algorithm in many parallel
threads. This puts more stress on the interconnect as any time the network used
to be idle can now be used by another thread. I collapsed the bandwidth reports
into a single metric reporting average bandwidth over the entire run.

### DIFF
--- a/gloo/benchmark/runner.cc
+++ b/gloo/benchmark/runner.cc
@@ -14,6 +14,7 @@
 
 #include "gloo/barrier_all_to_one.h"
 #include "gloo/broadcast_one_to_all.h"
+#include "gloo/common/common.h"
 #include "gloo/common/logging.h"
 #include "gloo/rendezvous/context.h"
 #include "gloo/transport/device.h"
@@ -56,6 +57,11 @@ Runner::Runner(const options& options) : options_(options) {
   }
 #endif
   GLOO_ENFORCE(device_, "Unknown transport: ", options_.transport);
+
+  // Spawn threads that run the actual benchmark loop
+  for (auto i = 0; i < options_.threads; i++) {
+    threads_.push_back(make_unique<RunnerThread>());
+  }
 
 #if GLOO_USE_REDIS
   if (!contextFactory_) {
@@ -166,6 +172,10 @@ void Runner::run(BenchmarkFn<T>& fn) {
 
 template <typename T>
 void Runner::run(BenchmarkFn<T>& fn, int n) {
+  std::vector<std::unique_ptr<Benchmark<T>>> benchmarks;
+
+  // Initialize one set of objects for every thread
+  for (auto i = 0; i < options_.threads; i++) {
     auto context = newContext();
     auto benchmark = fn(context);
     benchmark->initialize(n);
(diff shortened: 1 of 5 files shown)


gloo/benchmark/runner.cc
```diff
@@ -256,15 +302,15 @@ void Runner::printHeader() {
...
+  std::cout << std::setw(13) << ("avg " + bwSuffix);
   std::cout << std::setw(11) << "samples";
   std::cout << std::endl;
 }
 
-void Runner::printDistribution(int elements, int elemSize) {
+void Runner::printDistribution(
+    int elements,
+    int elementSize,
+    Distribution& samples) {
```
romain-intel: const ? It doesn't seem to be modified. Not sure how picky you want to be on const info.

gloo/benchmark/runner.cc
```diff
@@ -307,5 +346,44 @@ template void Runner::run(BenchmarkFn<float>& fn, int n);
...
+
+void RunnerThread::spawn() {
+  std::unique_lock<std::mutex> lock(mutex_);
+  while (!stop_) {
+    while (job_ == nullptr) {
+      cond_.wait(lock);
+      if (stop_) {
+        return;
+      }
+    }
```
romain-intel: Maybe copy job_ to a temporary location so you can release the lock here and allow for faster shutdown? This assumes that the job is only being run by one thread at a time (I think that's the case) since samples_ is not thread safe.

gloo/benchmark/runner.cc
```diff
@@ -187,37 +197,72 @@ void Runner::run(BenchmarkFn<T>& fn, int n) {
...
+      auto fn = [&benchmark = benchmarks[i]] { benchmark->run(); };
+      auto job = make_unique<RunnerJob>(fn, options_.warmupIterationCount);
+      jobs.push_back(std::move(job));
+    }
 
-      // Broadcast duration of fastest iteration during warmup,
-      // so all nodes agree on the number of iterations to run for.
-      auto nanos = broadcast(warmup.min());
-      iterations = std::max(1L, options_.iterationTimeNanos / nanos);
+    // Start jobs on every thread (synchronized across processes)
```
romain-intel: You do not seem to be "pinning" threads to anything so I am not sure how effective the "warmup" will be. It seems to be mostly just to agree on the iteration count so it should be OK. If you do want to use it to warmup the system, you may not be getting exactly what you want.

gloo/benchmark/runner.cc
```diff
@@ -187,37 +197,72 @@ void Runner::run(BenchmarkFn<T>& fn, int n) {
...
+  for (auto i = 0; i < options_.threads; i++) {
+    samples.merge(jobs[i]->getDistribution());
+  }
+
+  // Print results
+  printDistribution(n, sizeof(T), samples);
+
+  // Barrier to make sure everybody arrived here and the temporary
+  // context and benchmark can be destructed.
+  barrier_->run();
```
romain-intel: Isn't the wait above sufficient?

gloo/benchmark/runner.cc
pietern: It's fine if it's not super accurate. They will kick off at the same time, so the hope is they run in parallel and we get some good measurement of how long an iteration takes under the configured parallelism. Totally best effort.

gloo/benchmark/runner.cc
pietern: It should be, yeah. Trying it out :)

gloo/benchmark/runner.cc
pietern: Good call.

gloo/benchmark/runner.cc
pietern: Yeah, one at a time. The main thread is the one handing out jobs to run.

pietern:
@romain-intel Made the distribution of latency samples const correct and removed the superfluous barrier.

pietern:
Thanks @romain-intel!

PR #62, opened 2017-08-04
reorder if/else to fix cuda 7.5 bug

soumith:
^ what title says. Fixes CUDA 7.5 compilation.


pietern:
Nice catch, thanks @soumith.

PR #63, opened 2017-08-04
Gloo fixes for Linux + old cmake (2.8.0) + old glibc (CentOS6)

soumith:
let me know if you would like to have these changes.

Since gloo already assumes linux-only in some parts of the code, I used `memalign`. 
`aligned_alloc` is not present (unfortunately) in the glibc of centos6.

If you dont like these changes, I can create something more in the line of:
- check for aligned_alloc's existence during compile-time via cmake
- if available use it, else fallback to memalign

The reason I've ported the code back to the old cmake and old glibc is because when building binaries for PyTorch, we use CentOS6 (for the old glibc) so that the binaries run pretty much anywhere.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,4 +1,4 @@
-cmake_minimum_required(VERSION 2.8.12 FATAL_ERROR)
+cmake_minimum_required(VERSION 2.8 FATAL_ERROR)
 
 project(gloo CXX C)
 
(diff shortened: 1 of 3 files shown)


pietern:
Yeah, this is great. But only way to ensure this doesn't regress is by doing CI on this old stuff as well.

Let me check how to put that in place before merging.

pietern:
The memalign stuff is now merged. If you're still interested in the older CMake stuff, can you submit another PR for it? Otherwise we can ignore and expect 2.8.12, which I think is not unreasonable.

PR #64, opened 2017-08-04
Take CUDA peer access into account for on-device reduction

pietern:
The NVLink cube mesh architecture has partial peer access between devices. Two
groups of 4 GPUs have full peer access and every GPU in one group has peer
access to a corresponding GPU in the other group. When we reduce between all 8
GPUs, tree reduction using peer access must be done separately in those groups,
followed by a reduction across any one of the pairs connecting the groups. This
change refactors CudaDeviceReduce to work with this topology.

Additionally, it shuffles device pointers to randomize which GPUs run the
reduction and which communication links are used. The goal of randomization
here is to prevent excessive load on any single GPU or link.

### DIFF
--- a/gloo/cuda_collectives_device.h
+++ b/gloo/cuda_collectives_device.h
@@ -32,120 +32,228 @@ class CudaLocalDeviceReduce : public LocalOp<T> {
       size_t count)
       : streams_(streams),
         targetPtr_(targetPtr.range(offset, count)),
-        fn_(fn),
-        numPtrs_(devicePtrs.size()),
-        steps_(log2(numPtrs_)) {
-    // Only works with power-of-2 number of pointers
-    GLOO_ENFORCE(1 << steps_, streams.size(), "Not power of two");
-
+        fn_(fn) {
     // Incorporate offset/count into devicePtrs
     devicePtrs_.reserve(devicePtrs.size());
     for (const auto& ptr : devicePtrs) {
       devicePtrs_.push_back(ptr.range(offset, count));
     }
 
-    // Add level of indirection so that we can shuffle this instead
-    // of shuffling BOTH the streams and device ptr vectors.
-    for (auto i = 0; i < numPtrs_; i++) {
-      indices_.push_back(i);
-    }
+    // Initialize reduction sequence and setup peer access
+    computeIndices();
+    initializePeerAccess();
+  }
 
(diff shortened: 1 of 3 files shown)


pietern:
Fixing a few bugs...

gloo/cuda_collectives_device.h
```diff
@@ -32,120 +32,228 @@ class CudaLocalDeviceReduce : public LocalOp<T> {
...
-        // Queue reduction
-        fn_->call(
-            devicePtrs_[indexA],
-            devicePtrs_[indexB],
-            devicePtrs_[indexA].getCount(),
-            streamA);
+    // Shuffle order in an attempt to evenly spread work across devices when
+    // dealing with multiple instances of this operation.
+    auto& groupA = indices_[0];
+    std::random_shuffle(groupA.begin(), groupA.end());
```
wesolwsk: It may be better to explicitly set up non-overlapping communication topologies and iterate through them. Random may not work great because a single oversubscribed link may lead to 2x overall degradation. Can you make a TODO comment in the code?

gloo/cuda_collectives_device.h
```diff
@@ -32,120 +32,228 @@ class CudaLocalDeviceReduce : public LocalOp<T> {
...
-    // Wait for the final memory copy to complete
-    auto root = indices_[0];
-    streams_[root].wait();
+  void initializePeerAccess() {
+    // Initialize peer access within groups
+    for (const auto& group : indices_) {
+      int steps = log2(group.size());
+      GLOO_ENFORCE(
+          1 << steps,
+          group.size(),
```
soumith: one thing I dont see being taken into account is that peer access has a device limit of 8.
Even if deviceA and deviceB can be p2p enabled, after you register 8 devices for p2p, the subsequent `cudaDeviceEnablePeerAccess` calls will start getting rejected.
We particularly see this issue on 8 x K80 machines which see 16 visible devices.

Both Caffe2 and Torch work around this. See the Torch patch for example: https://github.com/torch/cutorch/pull/774

Hiroki11x:
Was this problem solved?

pietern:
@Hiroki11x Making a proper fix for this is still on the back burner. In the mean time you can compile with NCCL support and everything will work out of the box. Are you looking to specifically use the non-NCCL approach/code?

PR #74, opened 2017-08-22
BUILD_SHARED_LIBS is an on/off variable

### LINKED ISSUE #73: Fix use of BUILD_SHARED_LIBS
Right now it must be set to "SHARED" or "STATIC", but it should take "ON" or "OFF" like any other stock CMake variable.

pietern:
How it was used implied having to set it to STATIC or SHARED, which
is not compatible with the CMake predicated use of this variable.

Fixes #73.


pietern:
Verified that building with `-DBUILD_SHARED_LIBS=ON` generates a shared library.

PR #80, opened 2017-08-30
Stop relying on bundled version of NCCL

pietern:
NCCL1 is deprecated in favor of NCCL2 and if anyone wants to still use
it they can specify to do so via NCCL_ROOT_DIR.

This change removes the NCCL submodule and makes use of NCCL a compile
time option and constant. If USE_NCCL=ON, then all device side
reductions are performed using NCCL. Otherwise, they are performed with
a tree of CUDA reductions. Calls to cudaDeviceReduce or
cudaDeviceBroadcast are handled based on the USE_NCCL constant.


pietern:
Build fails because of submodule removal. I think it's safe to merge...

PR #90, opened 2017-09-26
Add AllreduceBuilder algorithm instance builder

pietern:
This builder class is an initial stab at moving the decision which
algorithm implementation to use to Gloo. Different machine or network
configurations, or buffer sizes, warrant different implementations.
Hardcoding these decisions over and over in downstream code is not
productive. Instead, we can (eventually) have Gloo make these
decisions for us.

Another benefit of deferring this decision is that we can more easily
choose the optimal implementation in case there is just a single
process with multiple buffers. Downstream code currently has to make a
distinction between single machine and multi machine configurations
and removing this distinction means these code paths can be
simplified.

This commit only contains a builder for Allreduce. The benchmark code
only uses the builder in case it runs an allreduce benchmark and the
current approach otherwise. When builders for the other collectives
are added we should up the benchmark main code as well.

### DIFF
--- a/gloo/CMakeLists.txt
+++ b/gloo/CMakeLists.txt
@@ -7,6 +7,7 @@ set(GLOO_HDRS)
 # Compiled sources in root directory
 list(APPEND GLOO_SRCS
   "${CMAKE_CURRENT_SOURCE_DIR}/algorithm.cc"
+  "${CMAKE_CURRENT_SOURCE_DIR}/allreduce_local.cc"
   "${CMAKE_CURRENT_SOURCE_DIR}/context.cc"
   )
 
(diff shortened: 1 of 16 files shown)


pietern:
@slayton58 Can you take a look and let me know what you think? Through this (or something like it) it will be easy to pile on dispatching logic.

slayton58:
That looks decent to me - it looks like the builder has full knowledge of the topology via the passed context?

pietern:
@slayton58 It should have all the information it needs and where it doesn't yet we can add it later without requiring changes to downstream code. This is also where you could switch in NCCL2 conditional on running on DGX-1 for example. We will need something richer than a switch/case to choose of course, but we can start with this pattern and then expand.

slayton58:
Great :) Is there consideration for running all possibilities and timing them in order to semi-automatically find the fastest (analogue to `cudnnFind*` routines)

pietern:
@slayton58 We can add that on top of this I think. As long as we have the inputs, the length of the input, and a context, we can do whatever we want to return an algorithm instance. So instead of `build(context)` you would call `find(context)` (or with some more params, or params you set on the builder), and it will iterate through the implementations and pick the best one. That said, if we do a micro benchmark in isolation, it may not be representative of the performance when there's N collectives running concurrently. Not a blocking issue of course, but noteworthy.

slayton58:
True enough, but I'm not sure that's a solvable problem in general for a library!

pietern:
Added test case.

pietern:
Added test cases for CUDA and split builder tests to separate files.

PR #91, opened 2017-09-27
Add XYZ_INCLUDE_DIR / XYZ_LIB_DIR for CMake dependencies

pietern:
It is flexible to have both XYZ_INCLUDE_DIR and XYZ_LIB_DIR for each dependency instead of just XYZ_ROOT_DIR. The gloo build should work with whatever directory structure is used for dependencies.

This change also slightly changes the CMake variable naming to remove some cognitive overhead regarding upcase/downcase variables and multiple messages in the CMake output.

Also see pytorch/pytorch#2877


soumith:
sweet, looks good to me.

PR #92, opened 2017-09-29
Added getDeviceNames helper function to get all IBVerb names

teng-li:
Added getDeviceNames as a helper function to get all the IB devices so that we can use it as a lib.

Tested by building 
cmake ../ -DBUILD_TEST=ON -DBUILD_BENCHMARK=ON -DUSE_MPI=ON -DUSE_CUDA=ON -DUSE_NCCL=ON -DUSE_IBVERBS=ON -DCMAKE_SKIP_BUILD_RPATH=TRUE -DNCCL_ROOT_DIR=$NCCL_ROOT_DIR

and 

Run IB benchmark with success

### DIFF
--- a/gloo/transport/ibverbs/device.cc
+++ b/gloo/transport/ibverbs/device.cc
@@ -15,7 +15,6 @@
 
 #include <algorithm>
 #include <array>
-#include <vector>
 
 #include "gloo/common/error.h"
 #include "gloo/common/linux.h"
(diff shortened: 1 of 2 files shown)


gloo/transport/ibverbs/device.cc
```diff
@@ -28,51 +26,18 @@ namespace ibverbs {
...
-
-  struct ibv_device*& operator[](int i) {
-    return list_[i];
-  }
-
- protected:
-  int size_;
-  struct ibv_device** list_;
-};
-
```
pietern: Can you leave this one in place? It's just a RAII helper, doesn't need to be in the header.

gloo/transport/ibverbs/device.h
```diff
@@ -32,6 +34,45 @@ struct attr {
...
+  int size() {
+    return size_;
+  }
+
+  struct ibv_device*& operator[](int i) {
+    return list_[i];
+  }
+
+  // Helper function that returns the list of IB device names in sorted order
+  std::vector<std::string> getDeviceNames() {
```
pietern: Please only add std::vector<std::string> getDeviceNames() without implementation to the header, then it can be implemented in device.cc, using the ibv_devices class.

teng-li:
@pietern Comments addressed

pietern:
Thanks @teng-li

PR #94, opened 2017-10-03
Add optional handling of MPI lifecycle to MPI context

pietern:
Applications that are launched through MPI but only use Gloo directly
shouldn't have to deal with MPI_Init/MPI_Finalize. By using the
createManaged() function on the MPI context, an application can now
defer dealing with MPI_Init and MPI_Finalize to Gloo.

### DIFF
--- a/gloo/mpi/context.cc
+++ b/gloo/mpi/context.cc
@@ -11,6 +11,7 @@
 
 #include <algorithm>
 #include <cstring>
+#include <mutex>
 
 #include "gloo/common/error.h"
 #include "gloo/common/logging.h"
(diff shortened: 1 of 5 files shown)


pietern:
@lukeyeager Can you take a look? I tested this only with the examples so far.

gloo/mpi/context.cc
```diff
@@ -33,6 +34,53 @@ static int MPICommRank(const MPI_Comm& comm) {
...
+      sptr = std::make_shared<MPIScope>();
+      wptr = sptr;
+    });
+
+  // Grab reference to MPIScope.
+  // This is a potential race since contexts may be destructed while
+  // new ones are constructed. No easy way to fix this.
+  if (!sptr) {
+    GLOO_ENFORCE(!wptr.expired(), "Cannot initialize MPIScope twice");
+    sptr = wptr.lock();
```
lukeyeager: Looks like you can address this race condition by checking whether the result of `lock()` is empty:
```cpp
    if (auto observe = weak.lock()) {
        std::cout << "\tobserve() able to lock weak_ptr<>, value=" << *observe << "\n";
    } else {
        std::cout << "\tobserve() unable to lock weak_ptr<>\n";
    }
```
http://en.cppreference.com/w/cpp/memory/weak_ptr/lock

pietern: Thanks, updated to only use `.lock()`.

lukeyeager:
Tried to run the tests but got hung up on https://github.com/facebookincubator/gloo/commit/c6247c450c1271a44870d0c02c721c087dc3a777#comments.

PR #97, opened 2017-10-05
Fix synchronization issue with cudaMemcpy

pietern:
Running two or more test processes in parallel would result in
spurious test failures. Only the synchronous CUDA tests would fail.
This was caused by the incorrect assumption that cudaMemcpy was
synchronous w.r.t. the host process. Instead, cudaMemcpy is launched
on the default stream (stream 0) of a device, and will be synchronous
only w.r.t. kernels launched on blocking streams on the same device.
The additional load on stream 0 caused by running multiple processes
resulted in compute kernels launched on non-blocking streams (as part
of the collective implementations) to run *before* the cudaMemcpy call
on stream 0 would finish, causing them to see invalid data.

This is fixed by adding explicit stream synchronization in
CudaFixture::assignValues and banning use of cudaMemcpy and implicit
synchronization in favor of cudaMemcpyAsync and explicit
synchronization.

Many thanks to @slayton58 and @lukeyeager for debugging.

### DIFF
--- a/gloo/test/cuda_base_test.h
+++ b/gloo/test/cuda_base_test.h
@@ -36,17 +36,30 @@ class CudaFixture : public Fixture<T> {
     }
   }
 
+  // Upon returning, the memory allocated by this fixture is set. The
+  // fixture data is visible to any kernel launched on any stream.
   void assignValues() {
     Fixture<T>::assignValues();
     for (auto i = 0; i < cudaSrcs.size(); i++) {
-      CUDA_CHECK(cudaMemcpy(
+      CudaDeviceScope scope(cudaStreams[i].getDeviceID());
+      CUDA_CHECK(cudaMemcpyAsync(
         *cudaSrcs[i],
         this->srcs[i].get(),
         cudaSrcs[i].bytes,
-        cudaMemcpyHostToDevice));
+        cudaMemcpyHostToDevice,
+        *cudaStreams[i]));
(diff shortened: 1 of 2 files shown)


gloo/test/cuda_base_test.h
```diff
@@ -36,17 +36,30 @@ class CudaFixture : public Fixture<T> {
...
         this->srcs[i].get(),
         cudaSrcs[i].bytes,
-        cudaMemcpyHostToDevice));
+        cudaMemcpyHostToDevice,
+        *cudaStreams[i]));
+    }
+    // Synchronize every stream to ensure the memory copies have completed.
+    for (auto i = 0; i < cudaSrcs.size(); i++) {
+      CudaDeviceScope scope(cudaStreams[i].getDeviceID());
+      CUDA_CHECK(cudaStreamSynchronize(*cudaStreams[i]));
```
lukeyeager: Consider this for the sake of DRY:
```cpp
void assignValues() {
  assignValuesAsync();
  synchronizeCudaStreams();
}
```

pietern: Yeah, looks nicer..

pietern: But then we have the cudaSleep in the middle!

pietern: Going with it as is

PR #102, opened 2017-11-09
Fixing CMakeLists.txt

Jorghi12:
Submitting a pull request for fixing the CMakeLists file.


pietern:
Looking at the difference between the original CMakeLists.txt and the current one, the comma's separating the builder headers need to be removed.

pietern:
There was an include missing in cuda.h after the std::all_of was moved there. Fixed in #105.

PR #106, opened 2017-11-14
Fix NCCL collectives with roots

slayton58:
Need to pass the NCCL rank as root instead of the root device ordinal.

Example of the previous issue: Imagine a 2 process, 4 GPU / process run with devices `0 : {0,1,2,3}, 1 : {4,5,6,7}`. Reduce on process 1 should pass `rank=0` whereas previously it would pass `rank=device[0]=4`, causing an error.


pietern:
Thanks @slayton58!

PR #108, opened 2017-11-15
NCCL2 multinode allreduce

slayton58:
This is based on work originally done by @lukeyeager, and is a first look at integration of the multi-node capabilities of NCCL2 into gloo as a new `Algorithm`.

Notes:
1. We cache created NCCL communicators in order to reduce both startup time and open-file limits that can be hit otherwise
2. Unless users pass streams, we create and re-use a set of streams for all reduction calls through NCCL2. This enforces an ordering of calls, and should alleviate deadlock concerns.
3. Currently the algorithm is synchronous with respect to the host - `cudaStreamSynchronize` is called at the end before returning. @pietern to comment.
4. Some re-arrangement of code needed to be done in `gloo/nccl/nccl.{h,cu}` in order to facilitate the work

### DIFF
--- a/gloo/cuda_allreduce_nccl2.h
+++ b/gloo/cuda_allreduce_nccl2.h
@@ -1,0 +1,76 @@
+/**
+ * Copyright (c) 2017-present, Facebook, Inc.
+ * All rights reserved.
+ *
+ * This source code is licensed under the BSD-style license found in the
+ * LICENSE file in the root directory of this source tree. An additional grant
+ * of patent rights can be found in the PATENTS file in the same directory.
+ */
+
+#pragma once
+
+#include "gloo/algorithm.h"
+#include "gloo/barrier_all_to_all.h"
+#include "gloo/cuda.h"
+#include "gloo/cuda_workspace.h"
+#include "gloo/nccl/nccl.h"
+
+#include <unordered_map>
+
+namespace gloo {
+
+class NCCLCommList {
+ public:
+  NCCLCommList(const std::shared_ptr<Context>& context,
+      const std::vector<int> localDevices);
+  ~NCCLCommList();
+  std::vector<ncclComm_t> comms;
+};
+
+class NCCLStreamList {
+ public:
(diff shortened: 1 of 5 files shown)


gloo/cuda_allreduce_nccl2.h
```diff
@@ -0,0 +1,78 @@
...
+
+  std::unique_ptr<LocalOp<T> > localReduceOp_;
+  std::unique_ptr<LocalOp<T> > localBroadcastOp_;
+
+  std::unique_ptr<transport::Buffer> sendDataBuf_;
+  std::unique_ptr<transport::Buffer> recvDataBuf_;
+
+  int dummy_;
+  std::unique_ptr<transport::Buffer> sendNotificationBuf_;
+  std::unique_ptr<transport::Buffer> recvNotificationBuf_;
```
pietern: A bunch of these are unused and should be removed

pietern:
This still has the single nccl comm per process instead of per context (so there is a risk of mismatching algorithms trying to use the same nccl comm, causing corruption). Let's wait with merging until that is in.

pietern:
Thanks @slayton58

slayton58:
@pietern Apologies about that - I'd pushed but hadn't actually added and commited the per-context change... Updated.

liamuk:
What's the current status of this?

pietern:
@liamuk This won't be merged anymore. In PyTorch, which is the primary user of this code, NCCL2 is exposed natively through c10d, and the same functionality is available (it's a better fit there in my opinion, since the context and transport abstractions don't apply for the NCCL2 algorithm here). See https://github.com/pytorch/pytorch/tree/master/torch/lib/c10d for the code there. The same features as this PR around stream and communicator reuse are applied there.

Are you using Gloo directly, or via PyTorch?

liamuk:
I'm just curious about the implementation details :) Was confused by this MR still being open, when the functionality was there in c10d.

pietern:
You're right, it can be closed.

PR #127, opened 2018-06-14
define SPEED_UNKNOWN for compilation with anaconda compilers

rainwoodman:
When compiling torch with anaconda compilers I ran into this error about SPEED_UNKNOWN undefined:
 
https://github.com/pytorch/pytorch/issues/4499

I couldn't find the proper header to include -- even if found it is likely not portable.
Given that not too many people outside kernel uses this macro (gloo is one of the few results googling with 'SPEED_KNOWN'), it is
probably a good idea to define it if it is not defined on the system.

Note that the comment on torch set the macro to -1.

The correct value appears to be 0, see e.g. this commit to the kernel:
https://patchwork.kernel.org/patch/9573661/

### DIFF
--- a/gloo/common/linux.cc
+++ b/gloo/common/linux.cc
@@ -32,7 +32,9 @@
 #include "gloo/common/logging.h"
 
 #ifndef SPEED_UNKNOWN
-/* some platforms doesn't define speed unknown in headers we included. */
+/* SPEED_UNKOWN is sometimes undefined, c.f.
+ * https://github.com/facebookincubator/gloo/pull/127
+ */
 #define SPEED_UNKNOWN 0
 #endif
 
--- a/gloo/common/linux.cc
+++ b/gloo/common/linux.cc
@@ -31,6 +31,11 @@
 
 #include "gloo/common/logging.h"
 
+#ifndef SPEED_UNKNOWN
+/* some platforms doesn't define speed unknown in headers we included. */
+#define SPEED_UNKNOWN 0
+#endif
+
 namespace gloo {
 
 const std::set<std::string>& kernelModules() {


pietern:
More specifically, which Linux kernel version. Looking at the history for the header that defines it I could only trace it back to version kernel version 3.7, but it may have been introduced even earlier.

rainwoodman:
The kernel is relatively recent:
```
kernel-4.16.11-100.fc26.x86_64
```
And I have installed `kernel-headers`.

I suspect the macro moved into a different header file or switched between internal and external at some point. On my system it is defined in `ethtool.h`.

pietern:
It's been in that header for at least the past 6 years, so perhaps something else is going on...

rainwoodman:
That's it. Anaconda uses cross compilation to ensure ABI compatibility with CentOS 5.x; therefore their ethtool.h is very old, as in ```./x86_64-conda_cos6-linux-gnu/sysroot/usr/include/linux/ethtool.h```,

```
/*
 * ethtool.h: Defines for Linux ethtool.
 *
 * Copyright (C) 1998 David S. Miller (davem@redhat.com)
 * Copyright 2001 Jeff Garzik <jgarzik@pobox.com>
 * Portions Copyright 2001 Sun Microsystems (thockin@sun.com)
 * Portions Copyright 2002 Intel (eli.kupermann@intel.com,
 *                                christopher.leech@intel.com,
 *                                scott.feldman@intel.com)
 * Portions Copyright (C) Sun Microsystems 2008
 */
```
SPEED_UNKNOWN is not defined in this version.

pietern:
That makes sense, thanks for figuring out. Merging the PR.

PR #130, opened 2018-08-23
Support for send/recv of ephemeral buffers

pietern:
The existing approach for the transports intends for buffers to be created once and used many times. Writes against these buffers are one sided. Coordination between sender and receiver (e.g. a clear to send signal) is typically done by using an auxiliary word-sized buffer that is signaled when it is OK to write.

This approach doesn't fit nicely with send/recv of ephemeral buffers that may be used only once. Additionally, there are no facilities to implement "receive from any". This PR adds the notion of an "unbound buffer" that does not tie to a particular pair or slot number. This buffer can be used as the source of a send operation or the destination of a receive operation. To coordinate between sender and receiver we introduce a set of OOB messages on the pair that can signal availability of either send/recv buffers. Currently this only provides an implementation for the TCP transport.


pietern:
@teng-li FYI, using this PR to push my WIP commits to. I'm keeping them stacked and also want to merge them as individual commits, so we can better bisect if any of them introduces a regression.

PR #138, opened 2018-10-09
Use type instead of memoryType when building with CUDA 10

syed-ahmed:
CUDA 10 builds of PyTorch are giving lots of __CUDA_DEPRECATED warnings with usage of `memoryType`. This PR should resolve those warnings in PyTorch builds.


pietern:
Thanks @syed-ahmed, LGTM.

PR #146, opened 2018-10-29
don't link against libcuda and libnvrtc

anderspapitto:


pietern:
@anderspapitto Can you also remove the find_library calls for libcuda and libnvrtc?

pietern:
Thanks @anderspapitto!

PR #164, opened 2019-01-25
Use single object to keep track of pending ops

pietern:
This used to be tracked in the pair instances, but to support
receive-from-any we need a centralized view into all pending send
operations. This class facilitaties that centralized view for both
pending send and receive operations.

Before this change, instances of the gloo::transport::tcp::Pair class
kept track of remote pending send and recv operations, and notified the
transport context of remote pending send operations. Functionality wise
this was fine, but it didn't include cleanup of unused slots, nor did it
allow for adding this easily. By centralizing the pending operation
tally in a single object, there is only a single object to clean up when
the pending operation tally drops to 0.

This resulted in a memory leak proportional to the number of used slots,
as reported in pytorch/pytorch#16144. Confirmed that the repro listed in
that issue no longer shows an increase in memory. Also added a test in
gloo/test/memory_test.cc that performs the same test.

### DIFF
--- a/gloo/transport/tcp/context.cc
+++ b/gloo/transport/tcp/context.cc
@@ -18,19 +18,63 @@ namespace gloo {
 namespace transport {
 namespace tcp {
 
+ContextMutator::ContextMutator(Context& context, size_t slot, size_t rank)
+    : lock_(context.m_),
+      context_(context),
+      slot_(slot),
+      rank_(rank),
+      it_(context_.remotePendingOp_.find(slot)) {}
+
+ContextMutator::~ContextMutator() {
+  if (it_ != context_.remotePendingOp_.end() && it_->second.empty()) {
+    context_.remotePendingOp_.erase(it_);
+  }
+}
+
+int8_t ContextMutator::getRemotePendingRecv() {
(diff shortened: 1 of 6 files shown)


gloo/transport/tcp/context.cc
```diff
@@ -105,13 +141,13 @@ int Context::recvFromAnyFindRank(
   return -1;
 }
 
-UnboundBuffer* Context::recvFromAnyCallback(
-    int rank,
+// Allowed to be called only by ContextMutator::findRecvFromAny.
+// The context lock is already held by the caller.
```
mrshenli: If I understood it correctly, it requires the lock has been acquired before calling `Pair::read()`, which is true for current use cases. Do we need to add `try_lock` checks for verification in case new calling path reaches this method? Or is it too costly to be enforced here?

pietern: The lock will have been acquired by the ContextMutator instance that will call this function.

pietern: I think enforcing this is costly though... It's a protected function, so only friend classes can call it, which is good enough I think.

PR #172, opened 2019-03-11
Add header files for std::array into source code

jcooky:
We use this library with clang and libc++. But, it does not compile with libc++.

image

When it compiles, this error occurred. Then, we send this PR for fix to add include files for std::array

We also add that includes for benchmark and google test
  - e.g. `cmake . -DBUILD_TEST=ON -DBUILD_BENCHMARK=ON`


pietern:
@jcooky Thank you for the contribution!

PR #176, opened 2019-04-26
Fix to adding include std::array in cuda file

cheonyujung:
I use this library with clang, libc++, cuda. But, it can't compile with libc++.

스크린샷 2019-04-26 오후 4 33 19

I think It's the same problem as https://github.com/facebookincubator/gloo/pull/172.
So I also fixed to add include files for std::array in cuda file.

I tested in clang 6, cuda 10.0.


pietern:
Thank you for fixing this issue!

PR #178, opened 2019-05-06
Add ReduceOp.XOR to torch.distributed

knottb:
Summary: Add ReduceOp.XOR to torch.distributed for use with torch.LongTensor allreduce functions and modify unit-tests accordingly

Differential Revision: D15223418


pietern:
Discussed with @knottb and this will be a change to PyTorch directly.

PR #194, opened 2019-08-01
Fix for Pytorch Integration

petrex:
The ${HIPIFY_OUTPUT_ROOT_DIR} was accidentally removed in the previous PR; however, we need this env variable for pytorch integration.


pietern:
Are you sure this fixes the issue? You can always submit a PR to PyTorch where you change the submodule URL to point to your fork to double check that it does.

petrex:
Yes, I got the signal from Pytorch CI. 

See this [PR](https://github.com/pytorch/pytorch/pull/23677).

pietern:
Thanks for confirming it works now, @petrex. Merging.

pietern:
The CI failures look like infrastructure problems.

PR #195, opened 2019-08-07
[WIP] Create libuv transport

pietern:
This transport is largely modeled after the existing TCP transport.
There is no support for the old style and only for the new style
algorithms (there are no createSendBuffer/createRecvBuffer functions).

Tests can be run by setting GLOO_TEST_TRANSPORT=uv. By default, none
of the tests against this transport run.

This commit adds the basic functionality but more work is needed to
properly deal with errors and timeouts.

It requires libuv to be available. Either it should already be defined
as a build target, if included by another project's build, or it must
be installed and discoverable through pkg-config.


gloo/transport/uv/pair.cc
```diff
@@ -0,0 +1,530 @@
...
+      // Read into unbound buffer, if the read is non-empty.
+      if (op.length) {
+        handle_->read((char*)op.buf->ptr + op.offset, op.length);
+        return;
+      }
+    }
+
+    GLOO_ENFORCE(op.nread == op.preamble.nbytes);
+    onSendUnboundBuffer(op);
+  } else if (opcode == Op::NOTIFY_SEND_READY) {
```
mrshenli: What is the protocol here? When do we need `NOTIFY_SEND_READY ` and `NOTIFY_RECV_READY ` ops? Is there always a handshake before sending real `SEND_UNBOUND_BUFFER ` ops? If so, why? If the receiver can accept `NOTIFY_SEND_READY `, can't it just directly accept `SEND_UNBOUND_BUFFER`?

gloo/transport/uv/pair.cc
```diff
@@ -0,0 +1,530 @@
...
+  localPendingRecv_[tag].emplace_back(std::move(buf), offset, nbytes);
+  sendNotifyRecvReady(tag, nbytes);
+
+  // Decrease balance of remote pending sends.
+  mutator.updateRemotePendingSend(-1);
+  return true;
+}
+
+void Pair::close() {
+  // abort();
```
mrshenli: Why do we need this comment here?

gloo/transport/uv/pair.cc
```diff
@@ -0,0 +1,530 @@
...
+// Receive into the specified buffer from the remote side of pair.
+void Pair::recv(
+    transport::UnboundBuffer* tbuf,
+    uint64_t tag,
+    size_t offset,
+    size_t nbytes) {
+  auto buf = static_cast<UnboundBuffer*>(tbuf)->getWeakNonOwningPtr();
+
+  GLOO_ENFORCE_GE(offset, 0);
+  GLOO_ENFORCE_GE(nbytes, 0);
```
mrshenli: can `offset` and `nbytes` ever become negative? If not can we directly use unsigned?

gloo/transport/uv/pair.cc
```diff
@@ -0,0 +1,530 @@
...
+
+  std::unique_lock<std::mutex> lock(mutex_);
+
+  // Notify peer of this pending recv.
+  localPendingRecv_[tag].emplace_back(std::move(buf), offset, nbytes);
+  sendNotifyRecvReady(tag, nbytes);
+
+  // Decrease balance of remote pending sends.
+  // This results in a negative count if executed before
+  // receiving the send notification from our peer.
```
mrshenli: Having some comments on the protocol will be very helpful.

gloo/transport/uv/unbound_buffer.h
```diff
@@ -0,0 +1,81 @@
...
+  std::condition_variable sendCv_;
+
+  int recvCompletions_;
+  int recvRank_;
+  int sendCompletions_;
+  int sendRank_;
+
+  // Allows for sharing weak (non owning) references to "this" without
+  // affecting the lifetime of this instance.
+  ShareableNonOwningPtr<UnboundBuffer> shareableNonOwningPtr_;
```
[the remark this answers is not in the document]
pietern: See the commit message in de2b7df74e3a36cd10926b9b13c2558db5656b91.

The tl;dr is -- the memory that the buffer objects point to is not managed by Gloo, so we don't have a way to prevent deallocation of reuse of that memory. What we can do though, is take a short lived lock on the unbound buffer that prevents it from being destructed. This is needed because we read into or write from this unmanaged memory from our I/O thread.

gloo/transport/uv/unbound_buffer.cc
```diff
@@ -0,0 +1,133 @@
...
+    *rank = sendRank_;
+  }
+}
+
+void UnboundBuffer::send(
+    int dstRank,
+    uint64_t slot,
+    size_t offset,
+    size_t nbytes) {
+  if (nbytes == UINT64_MAX) {
```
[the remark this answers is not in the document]
pietern: It does. But for clarity it should use `SIZE_MAX` or `std::numeric_limits<size_t>::max()` because we're dealing with `size_t` here. Will address and change this across the board.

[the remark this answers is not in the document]
pietern: I added #200 to my stack of diffs that addresses this.

I plan to rebase this PR once that stack is merged. They're all preparation diffs for compiling on macOS.

gloo/transport/uv/pair.h
```diff
@@ -0,0 +1,248 @@
...
+  const int rank_;
+
+  // Timeout for operations executed against this pair.
+  const std::chrono::milliseconds timeout_;
+
+  // Address that the remote side of this pair can connect to.
+  Address addr_;
+
+  // Handle of the connection, if the connection was made successfully.
+  std::shared_ptr<libuv::TCP> handle_;
```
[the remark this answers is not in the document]
pietern: There is one pair per rank in a single context. A pair represents the underlying connection between two ranks and all of the associated runtime state (for example all pending or in flight send and recv operations).

It's a `shared_ptr` because of how the libuv wrapper works. I modeled it after `uvw` where every libuv resource is encapsulated in a shared_ptr. This is done because it doesn't force you to maintain ownership of a resource. Instead, you can create one, register a few lambdas for callbacks, issue a request, and then discard it again. Ownership is then passed to libuv by leaking the shared pointer upon making the call and "unleaking" it when the callback is made. If you do want to keep it around yourself as well, you'll need shared ownership.

gloo/transport/uv/pair.cc
```diff
@@ -0,0 +1,530 @@
...
+    int rank,
+    std::chrono::milliseconds timeout)
+    : context_(context),
+      device_(device),
+      rank_(rank),
+      timeout_(timeout),
+      errno_(0) {
+  // The address for this pair is represented by the sockaddr of the listening
+  // socket and a unique identifier. The peer connects to the address and then
+  // writes the identifier to the new connection. The device instance The
```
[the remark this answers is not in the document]
pietern: I'll add some ASCII art describing how a pair of processes get connected.

gloo/transport/uv/pair.cc
```diff
@@ -0,0 +1,530 @@
...
+  // address a peer must connect to encapsulates both the sockaddr and an
+  // identifier for this pair.
+  addr_ = device_->nextAddress();
+}
+
+Pair::~Pair() {
+  std::unique_lock<std::mutex> lock(mutex_);
+
+  if (handle_) {
```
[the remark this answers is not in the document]
pietern: See discussion about `shared_ptr` above.

gloo/transport/uv/pair.cc
```diff
@@ -0,0 +1,530 @@
...
+      {
+        std::unique_lock<std::mutex> lock(mutex_);
+        closed = true;
+        cv_.notify_one();
+      }
+    });
+
+    // Wait for call to close handle to execute.
+    while (!closed) {
+      cv_.wait(lock);
```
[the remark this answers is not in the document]
pietern: Good call. Updating.

gloo/transport/uv/pair.cc
```diff
@@ -0,0 +1,530 @@
...
+  //
+  // NOTE(pietern): This can be split out to a separate function so
+  // that we first initiate all connections and then wait on all of
+  // them. This should make context initialization a bit faster. It
+  // requires a change to the base class though, so let's so it after
+  // this new transport has been merged.
+  //
+  std::unique_lock<std::mutex> lock(mutex_);
+  while (!handle_ && errno_ == 0) {
+    cv_.wait(lock);
```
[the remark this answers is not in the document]
pietern: Yes.

gloo/transport/uv/pair.cc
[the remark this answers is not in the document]
pietern: Is it possible to use `wait_for` without timeout?

gloo/transport/uv/pair.cc
```diff
@@ -0,0 +1,530 @@
...
+    throw ::gloo::IoException(GLOO_ERROR_MSG(
+        "Error connecting to ",
+        peer.str(),
+        ": ",
+        libuv::ErrorEvent(errno_).what()));
+  }
+}
+
+void Pair::connectCallback(
+    std::shared_ptr<libuv::TCP> handle,
```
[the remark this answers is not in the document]
pietern: Yep!

gloo/transport/uv/pair.cc
```diff
@@ -0,0 +1,530 @@
...
+    // after this destructor has returned, we must wait for
+    // the call to close the handle to have executed.
+    auto closed = false;
+
+    // Explicitly capture the handle to force a copy of the
+    // shared_ptr. The `handle_` member field may be reset by an event
+    // listener before this deferred call executes.
+    auto handle = handle_;
+    device_->defer([=, &closed] {
+      handle->close();
```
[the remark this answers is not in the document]
pietern: On the remote side of the connection, yes. Locally, it shouldn't. But I'll add some more code to confirm it actually doesn't. The documentation says that it's safe to deallocate the underlying structure as soon as the close **callback** has been called. Not when you simply call close.

gloo/transport/uv/pair.cc
pietern: This is done because the remote recv may not yet have been posted, OR it may be a recv from any operation. In either case, the unbound buffer is only sent when the pair knows there is a remote recv operation ready (i.e. it has received a `NOTIFY_RECV_READY` for that tag).

The `NOTIFY_SEND_READY` op is used to match against a recv from any on the other side. For example, if you have a single receiver and many senders (as is the case for RPC), the flow is as follows. The receiver side won't post any messages, just wait for `NOTIFY_SEND_READY` to arrive. The first one to arrive (again for the right tag and from the set of allowed ranks) will receive back a `NOTIFY_RECV_READY` message to ACK that it can send, and it will send the buffer. If then a few more `NOTIFY_SEND_READY` messages come in, the receiver can pick any of them to be the sender when the next recv from any call is made.

This mechanism is identical between the uv and tcp backends.

gloo/transport/uv/pair.cc
pietern: Good catch. Need to fix and update this.

gloo/transport/uv/pair.cc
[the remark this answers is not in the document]
mrshenli: Oh, I mean this sentence: "The device instance The address a peer must connect to encapsulates both the sockaddr and an identifier for this pair."

gloo/transport/uv/pair.cc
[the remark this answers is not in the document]
mrshenli: Sorry, I mean `cv_.wait(...)`

gloo/transport/uv/pair.cc
pietern: They can't. You're right. The checks are superfluous. I copied them from the tcp code, where they were used to initially guard against empty send/recv operations. Later we changed this to allow empty send/recv operations and now they're meaningless. I'll submit a separate PR to remove them from the tcp backend as well.

gloo/transport/uv/pair.cc
pietern: I agree. Some of it has been documented along the way, in elaborate commit messages, but that's not an adequate substitute for an actual single document describing how this works.

pietern: Created #201 to deal with that separately.

pietern:
Thanks for the elaborate review, @mrshenli!

You can skip the code in `context.{cc,h}` and `unbound_buffer.{cc,h}`. These are copied pretty much verbatim. The bulk of this PR is in `pair.cc` and `device.cc`. The libuv wrapping code is in `libuv.h`.

gloo/transport/uv/context.h
```diff
@@ -0,0 +1,244 @@
...
+#include <gloo/common/memory.h>
+#include <gloo/transport/context.h>
+
+//
+// READ THIS FIRST!
+//
+// This file is a copy of "gloo/transport/tcp/context.h".
+//
+// Any modifications should be made to that file as well. This file is
+// not a top level construct because it assumes the transport supports
```
mrshenli: Should we then extract these to some `UnboundedBufferContext` and put that as a top level abstract class? Don't have to be done in this PR.

gloo/transport/uv/context.h
```diff
@@ -0,0 +1,244 @@
...
+    // Update the count for the specified rank.
+    // If the count for this rank changes from zero to non-zero, it
+    // increments the non-zero counter. If it changes from non-zero to
+    // zero, it decrements the non-zero counter.
+    count_t update(size_t rank, count_t v) {
+      auto cur = count_[rank];
+      if (cur == 0) {
+        cur += v;
+        if (cur != 0) {
+          nonzero_++;
```
mrshenli: Just to confirm, this is protected by the lock in `ContextMutator`, right?

gloo/transport/uv/context.cc
```diff
@@ -0,0 +1,196 @@
...
+count_t ContextMutator::getRemotePendingSend() {
+  if (it_ == context_.remotePendingOp_.end()) {
+    return 0;
+  }
+  return it_->second.getSend(rank_);
+}
+
+count_t ContextMutator::updateRemotePendingRecv(count_t v) {
+  auto it = insertIfNotExists();
+  return it->second.updateRecv(rank_, v);
```
mrshenli: where did we use `lock_`?

gloo/transport/uv/pair.cc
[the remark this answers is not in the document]
pietern: Ah! Fixed! Also added a more elaborate description of the process to `Pair::connect`.

gloo/transport/uv/pair.cc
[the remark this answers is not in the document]
pietern: Done.

gloo/transport/uv/pair.cc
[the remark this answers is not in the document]
pietern: Rewrote this a bit to only allow the destructor to continue after the close callback has fired.

gloo/transport/uv/pair.cc
pietern: This is fixed in #206.

gloo/transport/uv/context.h
pietern: Yes, it can and should be pulled to the top level since there is nothing transport specific in it.

pietern: I created #207 to track.

gloo/transport/uv/context.h
pietern: Yes, correct.

gloo/transport/uv/context.cc
pietern: When the context mutator is constructed it acquires the context lock and when it is destructed it releases it. This means it can only exist if the lock is held.

pietern:
Closing in favor of resubmitted version at #212.

PR #196, opened 2019-08-08
Make aligned_allocator use posix_memalign

pietern:
This change makes the aligned allocator use posix_memalign over
memalign, which is a Linux specific call. This change also moves the
aligned allocator to a separate file per the realization that it is
used only in benchmark and test code.


pietern:
Replaced by the stack at #197.

PR #197, opened 2019-08-08
Make aligned_allocator use posix_memalign

pietern:
* **#197 Make aligned_allocator use posix_memalign**

This change makes the aligned allocator use posix_memalign over
memalign, which is a Linux specific call. This change also moves the
aligned allocator to a separate file per the realization that it is
used only in benchmark and test code.

Differential Revision: [D16801637](https://our.internmc.facebook.com/intern/diff/D16801637)

### DIFF
--- a/gloo/common/CMakeLists.txt
+++ b/gloo/common/CMakeLists.txt
@@ -4,6 +4,7 @@ set(GLOO_COMMON_SRCS
   )
 
 set(GLOO_COMMON_HDRS
+  "${CMAKE_CURRENT_SOURCE_DIR}/aligned_allocator.h"
   "${CMAKE_CURRENT_SOURCE_DIR}/common.h"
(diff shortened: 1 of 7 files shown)


gloo/common/CMakeLists.txt
```diff
@@ -4,6 +4,7 @@ set(GLOO_COMMON_SRCS
   )
 
 set(GLOO_COMMON_HDRS
+  "${CMAKE_CURRENT_SOURCE_DIR}/aligned_allocator.h"
```
mrshenli: If this is linux specific, should we also move this into an if-else clause like what #199 did for `linux.h`?

pietern: It's not Linux specific luckily, but `memalign` is.

PR #198, opened 2019-08-08
Export macro for every transport that gets compiled

pietern:
* **#198 Export macro for every transport that gets compiled**

Differential Revision: [D16801634](https://our.internmc.facebook.com/intern/diff/D16801634)


gloo/CMakeLists.txt
```diff
@@ -70,6 +70,20 @@ if(USE_ROCM)
...
 
+# Compile TCP transport only on Linux (it uses epoll(2))
+if(${CMAKE_SYSTEM_NAME} STREQUAL "Linux")
+  set(GLOO_HAVE_TRANSPORT_TCP 1)
+else()
+  set(GLOO_HAVE_TRANSPORT_TCP 0)
+endif()
+
+# Compile ibverbs transport if libibverbs is available
+if(USE_IBVERBS)
```
mrshenli: Do we have tests to cover these flags?

pietern: We don't...

PR #199, opened 2019-08-08
Make compilation of Linux source files conditional

pietern:
Stack from [ghstack](https://github.com/ezyang/ghstack):
* #206 Remove superfluous asserts
* #204 Fix C++14 extension warning when compiling for C++11
* #200 Use constexpr value for unspecified nbytes argument
* **#199 Make compilation of Linux source files conditional**
* #198 Export macro for every transport that gets compiled
* #197 Make aligned_allocator use posix_memalign

Differential Revision: [D16801686](https://our.internmc.facebook.com/intern/diff/D16801686)

PR #200, opened 2019-08-09
Use constexpr value for unspecified nbytes argument

pietern:
* **#200 Use constexpr value for unspecified nbytes argument**

Differential Revision: [D16801685](https://our.internmc.facebook.com/intern/diff/D16801685)

### DIFF
--- a/gloo/transport/tcp/unbound_buffer.h
+++ b/gloo/transport/tcp/unbound_buffer.h
@@ -38,10 +38,10 @@ class UnboundBuffer : public ::gloo::transport::UnboundBuffer {
   // If specified, the destination of this send is stored in the rank pointer.
   void waitSend(int* rank, std::chrono::milliseconds timeout) override;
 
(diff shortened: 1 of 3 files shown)


gloo/transport/tcp/unbound_buffer.h
```diff
@@ -38,10 +38,10 @@ class UnboundBuffer : public ::gloo::transport::UnboundBuffer {
   // If specified, the destination of this send is stored in the rank pointer.
   void waitSend(int* rank, std::chrono::milliseconds timeout) override;
 
-  void send(int dstRank, uint64_t slot, size_t offset, size_t nbytes = 0)
+  void send(int dstRank, uint64_t slot, size_t offset, size_t count)
```
mrshenli: nit: Why changing it to `count` here? Doesn't this create inconsistency with `gloo/transport/unbound_buffer.h` and `gloo/transport/tcp/unbound_buffer.cc`?

pietern: Great catch -- fixing.

PR #204, opened 2019-08-12
Fix C++14 extension warning when compiling for C++11

pietern:
* **#204 Fix C++14 extension warning when compiling for C++11**

Differential Revision: [D16801684](https://our.internmc.facebook.com/intern/diff/D16801684)

### DIFF
--- a/gloo/benchmark/runner.cc
+++ b/gloo/benchmark/runner.cc
@@ -249,7 +249,8 @@ void Runner::run(BenchmarkFn<T>& fn, size_t n) {
     // Create warmup jobs for every thread
     std::vector<std::unique_ptr<RunnerJob>> jobs;
     for (auto i = 0; i < options_.threads; i++) {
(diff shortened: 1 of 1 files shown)


gloo/benchmark/runner.cc
```diff
@@ -249,7 +249,8 @@ void Runner::run(BenchmarkFn<T>& fn, size_t n) {
     // Create warmup jobs for every thread
     std::vector<std::unique_ptr<RunnerJob>> jobs;
     for (auto i = 0; i < options_.threads; i++) {
-      auto fn = [&benchmark = benchmarks[i]] { benchmark->run(); };
+      auto& benchmark = benchmarks[i];
+      auto fn = [&benchmark] { benchmark->run(); };
```
mrshenli: does this make any differences on performance? Or just better readability?

pietern: No, it's to make the compiler stop emitting a warning that this is a C++14 extension...

PR #210, opened 2019-08-26
Enable optional thread sanitizer build

pietern:
* #234 Fix data race in uv transport shutdown
* #211 Fix lock inversion issue
* **#210 Enable optional thread sanitizer build**

The suppressions file contains false positives and, if used, makes the
test run finish free of tsan errors.

To build with the thread sanitizer enabled:

```
cmake ../ -DBUILD_TEST=1 -DSANITIZE_THREAD=1 -DCMAKE_BUILD_TYPE=DEBUG
```

To use the suppressions file:

```
export TSAN_OPTIONS="suppressions=[PATH]/.tsan-suppressions"
```

Differential Revision: [D17258598](https://our.internmc.facebook.com/intern/diff/D17258598)


.tsan-suppressions
```diff
@@ -0,0 +1,4 @@
+race:gloo::transport::tcp::Pair::read
```
mrshenli: do we know why thread sanitizer thinks there are races here?

pietern: Yes, this is caused by the multi threaded tests where a send may not be protected by a mutex, whereas a read (triggered from the I/O thread) is always protected by a mutex.

PR #211, opened 2019-08-26
Fix lock inversion issue

pietern:
* #234 Fix data race in uv transport shutdown
* **#211 Fix lock inversion issue**
* #210 Enable optional thread sanitizer build

This path is executed when an error happens waiting for a send or recv
operation against an unbound buffer to complete. The unbound buffer
code then signals the context that an error has happened, which in
turn signals all the pair objects that an error has happened.

This is done such that all future operations will fail as well (as we
cannot guarantee the connected context is still usable after an I/O
error or a timeout).

The problem was rooted in the context lock being acquired before each
of the pair locks were, whereas during nominal operation, the pair
lock is acquired before the context lock.

The solution is to remove the context lock entirely, because the
`pairs_` vector is logically const anyway. It is constructed once and
destructed at context destruction. In between, we do not need
exclusion to access it.

Differential Revision: [D17258597](https://our.internmc.facebook.com/intern/diff/D17258597)


gloo/transport/tcp/context.cc
```diff
@@ -192,7 +192,10 @@ bool Context::findRecvFromAny(
 }
 
 void Context::signalException(const std::string& msg) {
-  std::unique_lock<std::mutex> lock(m_);
+  // The `pairs_` vector is logically constant. After the context and
+  // all of its pairs have been created it is not mutated until the
+  // context is destructed. Therefore, we don't need to acquire this
+  // context's instance lock before looping over `pairs_`.
   for (auto& pair : pairs_) {
```
mrshenli: Will there be a problem if `createPair` has not been called on all ranks? Will those late ranks miss the exception signal?

pietern: The creation of these pairs is done before any operation is run against them, so this can't happen. In order to protect against it, keeping the mutex around wouldn't be enough either, but would require adding some kind of state variable that tells us whether or not initialization is complete.

PR #225, opened 2019-10-16
Only set c++11 if not already set to c++14

smessmer:
This allows projects using this to specify c++14 or c++17 for their cuda files without it conflicting here.

Test Plan: My PR switching PyTorch to C++14 (https://github.com/pytorch/pytorch/pull/27864) has a lot of errors right now but fewer errors with this change (i.e. the gloo dependency builds correctly after this change).

### DIFF
--- a/cmake/Cuda.cmake
+++ b/cmake/Cuda.cmake
@@ -165,7 +165,7 @@ foreach(diag cc_clobber_ignored integer_sign_change useless_using_declaration se
   gloo_list_append_if_unique(CUDA_NVCC_FLAGS -Xcudafe --diag_suppress=${diag})
 endforeach()
 
-# Set C++11 support
+# If the project including us doesn't set any -std=xxx directly, we set it to C++11 here.
 set(CUDA_PROPAGATE_HOST_FLAGS OFF)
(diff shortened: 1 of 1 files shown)


cmake/Cuda.cmake
```diff
@@ -167,7 +167,9 @@ endforeach()
 
 # Set C++11 support
```
pritamdamania87: Need to fix this comment?

cmake/Cuda.cmake
```diff
@@ -167,7 +167,9 @@ endforeach()
...
 set(CUDA_PROPAGATE_HOST_FLAGS OFF)
-gloo_list_append_if_unique(CUDA_NVCC_FLAGS "-std=c++11")
+if((NOT "${CUDA_NVCC_FLAGS}" MATCHES "-std=c\\+\\+") AND (NOT "${CUDA_NVCC_FLAGS}" MATCHES "-std=gnu\\+\\+"))
```
pritamdamania87: Can we add a comment here explaining why we are doing this?

smessmer: I added an explanation to the comment above

pietern:
Looks good to me. Thanks @smessmer and @pritamdamania87!

PR #228, opened 2019-10-30
Add abortWaitSend/abortWaitRecv to unbound_buffer

### LINKED ISSUE #227: Interrupt waitSend/waitRecv
These calls on the `gloo::transport::UnboundBuffer` class can currently not be interrupted. They wait on a condition variable that is to be notified by the I/O thread, or they timeout. There should also be a way to interrupt them from the outside in, for example if you're terminating a process and want to stop waiting for new messages to arrive.

rohan-varma:
Summary: Closes #227.
Adds 2 methods: `abortWaitSend`, `abortWaitRecv` so that we can interrupt a blocking `waitRecv` or `waitSend`, per @pietern 's suggestion. This is useful if a gloo process is waiting for a message in one thread, but another thread decides to shutdown and wants to stop it gracefully.

The methods set a boolean variable and notify a waiting condition variable. The condition variable predicate is changed to check this condition. The implementation is both for the tcp and uv transport methods. Unit tests are also added.

Differential Revision: D18210908

### DIFF
--- a/gloo/test/send_recv_test.cc
+++ b/gloo/test/send_recv_test.cc
@@ -235,6 +235,79 @@ TEST_P(SendRecvTest, RecvFromAny) {
   });
 }
 
+TEST_P(SendRecvTest, AbortRecv) {
+  const auto transport = std::get<0>(GetParam());
+  const auto contextSize = std::get<1>(GetParam());
+  constexpr uint64_t slot = 0x1337;
+  spawn(transport, contextSize, [&](std::shared_ptr<Context> context) {
+    // Test that an unbounded buffer should be able to abort its own waitRecv.
+    if (context->rank == 0) {
+      bool recvCompleted;
+      int srcRank = -1;
+      int tmp;
+      auto buf = context->createUnboundBuffer(&tmp, sizeof(tmp));
+
+      std::thread waitRecvThreadAbort(
+          [&]() { recvCompleted = buf->waitRecv(&srcRank); });
+      buf->abortWaitRecv();
+      waitRecvThreadAbort.join();
+      ASSERT_FALSE(recvCompleted);
+
+      // Future waitRecvs should not be aborted if we received one abortWaitRecv
+      // previously
+      std::thread waitRecvThread([&]() {
+        std::vector<int> ranks = {1};
+        buf->recv(ranks, slot);
+        recvCompleted = buf->waitRecv(&srcRank);
+      });
+      waitRecvThread.join();
+      ASSERT_TRUE(recvCompleted);
+    } else if (context->rank == 1) {
+      // Send to rank 0
+      int tmp = context->rank;
+      auto buf = context->createUnboundBuffer(&tmp, sizeof(tmp));
+      buf->send(0, slot);
+      buf->waitSend();
+    }
+  });
+}
+
+TEST_P(SendRecvTest, AbortSend) {
+  const auto transport = std::get<0>(GetParam());
+  const auto contextSize = std::get<1>(GetParam());
+  constexpr uint64_t slot = 0x1337;
+  spawn(transport, contextSize, [&](std::shared_ptr<Context> context) {
+    // Test that an unbounded buffer should be able to abort its own waitSend
+    bool sendCompleted;
+    if (context->rank == 0) {
+      int tmp;
+      auto buf = context->createUnboundBuffer(&tmp, sizeof(tmp));
+
+      std::thread waitSendThreadAbort(
+          [&]() { sendCompleted = buf->waitSend(); });
+      buf->abortWaitSend();
+      waitSendThreadAbort.join();
+      ASSERT_FALSE(sendCompleted);
+
+      // Future wait sends should not be aborted if we received one abortSend
+      // previously
+      std::thread waitSendThread([&]() {
+        // Send to rank 1
+        buf->send(1, slot);
+        sendCompleted = buf->waitSend();
+      });
+      waitSendThread.join();
+      ASSERT_TRUE(sendCompleted);
+    } else if (context->rank == 1) {
+      // Receive from rank 0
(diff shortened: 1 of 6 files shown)


gloo/transport/tcp/unbound_buffer.cc
```diff
@@ -72,11 +84,12 @@ void UnboundBuffer::waitRecv(int* rank, std::chrono::milliseconds timeout) {
                   "ms for recv operation to complete"));
     }
   }
-
+  auto recvCompleted = recvCompletions_ > 0;
```
xush6528: Can't understand this state, `recvCompleted`. Mind adding comments to explain it a bit?

gloo/test/send_recv_test.cc
```diff
@@ -235,6 +235,48 @@ TEST_P(SendRecvTest, RecvFromAny) {
...
+    bool recvCompleted;
+    if (context->rank == 0) {
+      int tmp;
+      auto buf = context->createUnboundBuffer(&tmp, sizeof(tmp));
+
+      std::thread waitRecvThread([&]() {
+        std::vector<int> ranks = {1};
+        int srcRank = -1;
+        buf->recv(ranks, slot);
+        recvCompleted = buf->waitRecv(&srcRank);
```
xush6528: Not quite understand the concepts from gloo.

What's the purpose of calling `buf->recv(..)` first?

rohan-varma: From what I understand, `buf->recv` will call the `recv` function from a particular `pair` (basically an encapsulation of 2 gloo processes that can communicate), which notifies the sender that there is a receive pending, but does not block. Then, `waitRecv` blocks until `Pair::read` indicates that data has been received.

gloo/transport/tcp/unbound_buffer.cc
rohan-varma: Sure, I'll add some comments.

The goal here is to return `true` if the receive completed (i.e. we got data from the remote side) and `false` if it was aborted. One way to know that we got data from the remote side is if there was at least one `recvCompletion`, so I used this to determine the return value. We could potentially also use `!abortWaitRecv`, but we could run into an issue where the receive completes successfully, then a thread sets `abortWaitRecv` before we return out of `waitRecv`, in which case returning `!abortWaitRecv` would not be correct

pietern: If it was aborted, it should return here and not try and decrement `recvCompletions_`. This counter is a tally that counts the number of recv's that completed without an accompanying `waitRecv` call. If it is non-zero when `waitRecv` is called, `waitRecv` can return immediately without blocking (as evidenced by the condition `if (recvCompletions_ == 0)`).

gloo/transport/tcp/unbound_buffer.cc
```diff
@@ -72,11 +84,12 @@ void UnboundBuffer::waitRecv(int* rank, std::chrono::milliseconds timeout) {
...
   recvCompletions_--;
   if (rank != nullptr) {
     *rank = recvRank_;
   }
+  return recvCompleted;
```
pietern: If we have acknowledged this completion, we must always return true.

gloo/test/send_recv_test.cc
xush6528: I see, receive operation needs a signal first.
And it's similar code to `ProcessGroupGloo::recv` or `ProcessGroupGloo::recvAnysource`.

For normal recv, it's necessary. Otherwise, `RecvWork` will wait for nothing and block forever.

Since you are aborting anyway, I think this line could be removed to make the test subject prominent.

gloo/transport/tcp/unbound_buffer.cc
xush6528: Will this happen? 
Before it returns out of `waitRecv`, doesn't it hold the lock?

rohan-varma: That makes sense, the lock would be reacquired once the thread is awakened. This should be changed to

`if (abortWaitSend_): return false`
else, decrement `recvCompletions` as below and return true

rohan-varma:
@pietern @xush6528 Updated with the suggestions.

gloo/test/send_recv_test.cc
```diff
@@ -235,6 +235,47 @@ TEST_P(SendRecvTest, RecvFromAny) {
...
+  constexpr uint64_t slot = 0x1337;
+  spawn(transport, contextSize, [&](std::shared_ptr<Context> context) {
+    // test that an unbounded buffer should be able to abort its own waitRecv.
+    bool recvCompleted;
+    if (context->rank == 0) {
+      int tmp;
+      auto buf = context->createUnboundBuffer(&tmp, sizeof(tmp));
+
+      std::thread waitRecvThread([&]() {
+        std::vector<int> ranks = {1};
```
xush6528: This is not used.

rohan-varma:
>Right now, all future calls to waitSend and waitRecv will return false if abort has been called once. But the intent is more likely to just interrupt a single call, so we should toggle the abort flag back to `false` if it was `true` once.

That is a good catch. We should flip the flag like you mentioned

gloo/transport/tcp/unbound_buffer.cc
```diff
@@ -120,10 +135,14 @@ void UnboundBuffer::waitSend(int* rank, std::chrono::milliseconds timeout) {
     }
   }
 
+  if (abortWaitSend_) {
+    return false;
```
xush6528: So flip it back here.

```cpp
abortWaitSend_ = false;
```
Also need a test case for testing an "abort" call only cancles one "wait" call?

pietern:
@rohan-varma What was the issue in the libuv code?

xush6528:
Redo in #232.

PR #229, opened 2019-11-05
[do not merge] example for bad libuv changes

rohan-varma:
- This is a compiler error but still passes the current CI. We should add CI to test on macOS/with the `USE_LIBUV` flag with libuv available to catch these issues.


pietern:
I just pushed a merge commit that pulls in #230 to make sure this now fails.

pietern:
It fails as expected:

```
cd /root/project/build/gloo && /usr/bin/g++-5   -I/root/project -I/root/project/build -isystem /usr/local/include  -std=c++11 -fPIC -O3 -DNDEBUG   -o CMakeFiles/gloo.dir/transport/uv/unbound_buffer.cc.o -c /root/project/gloo/transport/uv/unbound_buffer.cc
/root/project/gloo/transport/uv/unbound_buffer.cc: In member function 'void gloo::transport::uv::UnboundBuffer::handleRecvCompletion(int)':
/root/project/gloo/transport/uv/unbound_buffer.cc:36:36: error: 'mutex_variable_doesnt_exist_' was not declared in this scope
   std::lock_guard<std::mutex> lock(mutex_variable_doesnt_exist_);
```

PR #230, opened 2019-11-05
Start using CircleCI

pietern:
This change swaps out the older Travis configuration for a new CircleCI configuration.

Tested configurations (all on Ubuntu 18.04):
* GCC 5
* GCC 7
* Clang 6
* GCC 7 with CUDA 9.2
* GCC 7 with CUDA 10.1

Every configuration builds all transports (tcp, ibverbs, uv).


pietern:
File issues for:
* Build benchmark
* Run benchmark?
* Build on macOS

PR #232, opened 2019-11-05
Add abortWaitSend/abortWaitRecv to unbound_buffer

rohan-varma:
Summary:
Closes https://github.com/facebookincubator/gloo/issues/227.
Adds 2 methods: abortWaitSend, abortWaitRecv so that we can interrupt a blocking waitRecv or waitSend, per pietern 's suggestion. This is useful if a gloo process is waiting for a message in one thread, but another thread decides to shutdown and wants to stop it gracefully.
The methods set a boolean variable and notify a waiting condition variable. The condition variable predicate is changed to check this condition. The implementation is both for the tcp and uv transport methods. Unit tests are also added.

Differential Revision: D18321454

### DIFF
--- a/gloo/test/send_recv_test.cc
+++ b/gloo/test/send_recv_test.cc
@@ -235,6 +235,80 @@ TEST_P(SendRecvTest, RecvFromAny) {
   });
 }
 
+TEST_P(SendRecvTest, AbortRecv) {
+  const auto transport = std::get<0>(GetParam());
+  const auto contextSize = std::get<1>(GetParam());
+  constexpr uint64_t slot = 0x1337;
+  spawn(transport, contextSize, [&](std::shared_ptr<Context> context) {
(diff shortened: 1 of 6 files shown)


xush6528:
```
terminate called after throwing an instance of 'gloo::EnforceNotMet'
  what():  [enforce fail at /root/project/gloo/transport/uv/pair.cc:234] op.buf. Cannot lock pointer to unbound buffer
Received 'aborted' signal
```

Error is relevant.

rohan-varma:
> ```
> terminate called after throwing an instance of 'gloo::EnforceNotMet'
>   what():  [enforce fail at /root/project/gloo/transport/uv/pair.cc:234] op.buf. Cannot lock pointer to unbound buffer
> Received 'aborted' signal
> ```
> 
> Error is relevant.

It looks like this has been fixed by adding `buf->waitRecv();` to the `AbortRecv` test.

PR #237, opened 2019-12-10
Move event loop to loop.{cc,h}

pietern:
* #243 Use a single listening socket per device
* #242 Add error class
* #241 Add RAII wrapper for socket
* #240 Allow deferring functions to the epoll(2) thread
* #239 Add sequence number to gloo::transport::tcp::Address
* #238 Move attr struct to own header
* **#237 Move event loop to loop.{cc,h}**
* #236 Create and use epoll(2) handler base class

Differential Revision: [D18909100](https://our.internmc.facebook.com/intern/diff/D18909100)

### DIFF
--- a/gloo/transport/tcp/loop.cc
+++ b/gloo/transport/tcp/loop.cc
@@ -20,7 +20,7 @@ namespace gloo {
 namespace transport {
 namespace tcp {
 
-Loop::Loop() : fd_(-1), done_(false) {
+Loop::Loop() {
   fd_ = epoll_create(1);
   GLOO_ENFORCE_NE(fd_, -1, "epoll_create: ", strerror(errno));
   loop_.reset(new std::thread(&Loop::run, this));
--- a/gloo/transport/tcp/loop.cc
+++ b/gloo/transport/tcp/loop.cc
@@ -1,0 +1,91 @@
+/**
+ * Copyright (c) 2017-present, Facebook, Inc.
+ * All rights reserved.
+ *
+ * This source code is licensed under the BSD-style license found in the
+ * LICENSE file in the root directory of this source tree.
+ */
+
+#include <gloo/transport/tcp/loop.h>
+
+#include <string.h>
+#include <unistd.h>
+
+#include <array>
+
+#include <gloo/common/error.h>
+#include <gloo/common/logging.h>
+
+namespace gloo {
+namespace transport {
+namespace tcp {
+
+Loop::Loop() : fd_(-1), done_(false) {
+  fd_ = epoll_create(1);
+  GLOO_ENFORCE_NE(fd_, -1, "epoll_create: ", strerror(errno));
+  loop_.reset(new std::thread(&Loop::run, this));
+}
+
+Loop::~Loop() {
+  if (loop_) {
+    done_ = true;
+    loop_->join();
+  }
+  if (fd_ >= 0) {
+    close(fd_);
+  }
+}
+
+void Loop::registerDescriptor(int fd, int events, Handler* h) {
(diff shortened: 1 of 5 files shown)


gloo/transport/tcp/loop.cc
```diff
@@ -0,0 +1,91 @@
...
+  GLOO_ENFORCE_NE(rv, -1, "epoll_ctl: ", strerror(errno));
+}
+
+void Loop::unregisterDescriptor(int fd) {
+  auto rv = epoll_ctl(fd_, EPOLL_CTL_DEL, fd, nullptr);
+  GLOO_ENFORCE_NE(rv, -1, "epoll_ctl: ", strerror(errno));
+
+  // Wait for loop to tick before returning, to make sure the handler
+  // for this fd is not called once this function returns.
+  if (std::this_thread::get_id() != loop_->get_id()) {
```
jjlilley: Should this be a ==, rather than != ?

Do you need to 'kick' the handler to force an event here, so that we don't wait forever? (Is the blocking time an issue in practice)

fwiw, a common alternative non-blocking approach is to use explicit refcounting on the Handler object, to ensure the handler lifetime (e.g. registerFd is +1, getting an event is a +1 while event processing happens). I realize that this is a different model than you have here that might not be suited to the requirements... I personally developed an affinity for refcounts over explicit arbitrary blocking, but wanted to bring this up.

gloo/transport/tcp/loop.cc
```diff
@@ -0,0 +1,91 @@
...
+void Loop::run() {
+  std::array<struct epoll_event, capacity_> events;
+  int nfds;
+
+  while (!done_) {
+    // Wakeup everyone waiting for a loop tick to finish.
+    cv_.notify_all();
+
+    // Wait for something to happen
+    nfds = epoll_wait(fd_, events.data(), events.size(), 10);
```
jjlilley: Checking: do we really want to busy spin here every 10msec?
It's ok, though it feels like a bit of a let-down. We eventually could use a pair signal fd to avoid it, or... something? It seems reasonable to check in, but still TODO? (maybe I'm being silly?)

gloo/transport/tcp/loop.cc
pietern: This is a very good point, thanks for raising it. I agree doing refcounting on the handler is a better approach here. The loop can keep a weak ptr so if the underlying object goes away, it's all good (closing would be implicit unregistration of an fd). This is a bit more invasive change than I'd like to do here though, so I filed #246 to track this.

gloo/transport/tcp/loop.cc
pietern: I agree, this should be updated. I filed #246 to track.

gloo/transport/tcp/loop.cc
jjlilley: Sure, it seems quite reasonable to defer the changes to another PR - I'm fine with this going in as-is.

For the future, if we do have Handler in a shared_ptr<> at higher layers, it might be quite nice to just keep as shared_ptr<> down here in the Loop (barring refcount cycles, etc), so that we have handler lifetime guarantees and can avoid this waiting. It might make the tsan suppression in the other change go away as well :)

pietern:
@jjlilley Thanks for the approval.

One more thought on the refcounted handler issue: doing this effectively means anything that deals with I/O must be a shared_ptr as well, which could have some effect on the public facing APIs (e.g. like the pair instances on the context, which are currently unique_ptr's). I think this would make switching to shared_ptr handlers more involved than on the face of it.

PR #241, opened 2019-12-10
Add RAII wrapper for socket

pietern:
* (to be filled)

### DIFF
--- a/gloo/transport/tcp/socket.cc
+++ b/gloo/transport/tcp/socket.cc
@@ -1,0 +1,143 @@
+/**
+ * Copyright (c) 2017-present, Facebook, Inc.
+ * All rights reserved.
+ *
+ * This source code is licensed under the BSD-style license found in the
+ * LICENSE file in the root directory of this source tree.
+ */
+
+#include <gloo/transport/tcp/socket.h>
+
+#include <netinet/tcp.h>
+#include <string.h>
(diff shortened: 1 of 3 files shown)


gloo/transport/tcp/socket.cc
```diff
@@ -0,0 +1,175 @@
...
+  if (ss.ss_family == AF_INET6) {
+    const struct sockaddr_in6* sa = (const struct sockaddr_in6*)&ss;
+    return connect((const struct sockaddr*)sa, sizeof(*sa));
+  }
+  GLOO_ENFORCE(false, "Unknown address family: ", ss.ss_family);
+}
+
+void Socket::connect(const struct sockaddr* addr, socklen_t addrlen) {
+  auto rv = ::connect(fd_, addr, addrlen);
+  if (rv == -1 && errno != EINPROGRESS) {
```
jjlilley: Since failure is assert here, not sure if we want to handle EINTR? EAGAIN?

pietern: Good point. Updating to handle EINTR. I don't want to deal with EAGAIN on `connect(2)`, because it means something is wrong (or at least poorly tuned):

> EAGAIN: Insufficient entries in the routing cache.

jjlilley: SGTM

PR #242, opened 2019-12-10
Add error class

pietern:
* (to be filled)


gloo/transport/tcp/error.cc
```diff
@@ -0,0 +1,46 @@
...
+namespace tcp {
+
+const Error Error::kSuccess = Error();
+
+std::string Error::what() const {
+  return "no error";
+}
+
+std::string SystemError::what() const {
+  std::stringstream ss;
```
jjlilley: nit: maybe std::ostringstream for these?

(since this is error path, it shouldn't actually matter, but std::streamstream effectively has to pay the unused std::istreamstream constructor/destructor cost, which actually was showing up in my serialization hot-path profiles the other month)

pietern: Good point! Updated to use `std::ostringstream`.

PR #243, opened 2019-12-10
Use everything new!

pietern:
* (to be filled)


gloo/transport/tcp/helpers.h
```diff
@@ -0,0 +1,171 @@
...
+    auto leak = std::move(this->leak_);
+
+    // Write T.
+    auto rv = socket_->write(&t_, sizeof(t_));
+    if (rv == -1) {
+      fn_(socket_, SystemError("write", errno));
+      return;
+    }
+
+    // Check for short read (assume we can write in a single call).
```
jjlilley: You mean a short write/ ShortWriteError?

I don't know the typical sizeof(t_) or what's typical here, but does it make sense to not accept a short write here, and instead schedule another write?

If the process is keeping the kernel buffers full, it might be common to have short writes, and isn't necessarily such a bad situation, unless you need to guarantee that sizeof(t_) goes out.

pietern: Yes, good catch. This was a typical copy-paste error...

These helpers are used to send the sequence number around, which is only 8 bytes. Once that's done, the pair takes over control of the file descriptor and properly takes care of the partial reads/writes. Therefore I think it's fine to make this assumption here. The function is only called if `epoll(2)` says the socket is readable or writable, so we should be able to complete an 8 byte read/write.

jjlilley: 8 bytes should be quite benign then - I just didn't have context on the expected size here, wasn't sure if it was std::array<uint8_t, SomeBigN> or something :)

gloo/transport/tcp/listener.h
```diff
@@ -0,0 +1,74 @@
...
+// corresponding pair. If it can't be matched, it is stashed until a
+// pair with the sequence number calls `waitForConnection`.
+class Listener final : public Handler {
+ public:
+  using connect_callback_t =
+      std::function<void(std::shared_ptr<Socket> socket, Error error)>;
+
+  static constexpr auto kBacklog = 512;
+
+  explicit Listener(std::shared_ptr<Loop> loop, const attr& attr);
```
jjlilley: nit: do we use explicit on multi-arg constructors?

gloo/transport/tcp/listener.h
```diff
@@ -0,0 +1,74 @@
...
+
+  virtual ~Listener();
```
jjlilley: nit: if override is present, I think it's supposed to be used consistently

gloo/transport/tcp/listener.cc
```diff
@@ -0,0 +1,123 @@
...
+}
+
+Listener::~Listener() {
+  if (listener_) {
+    loop_->unregisterDescriptor(listener_->fd());
+  }
+}
+
+void Listener::handleEvents(int /* unused */) {
+  std::lock_guard<std::mutex> guard(mutex_);
```
jjlilley: This is fine for now, but I'm wondering if the code is locking more than it should, as it calls into syscalls and handlers with the lock.

gloo/transport/tcp/listener.h
pietern: Good catch. Fixed.

gloo/transport/tcp/listener.h
pietern: Indeed. Fixed.

gloo/transport/tcp/listener.cc
pietern: It is probably possible to make this more fine grained, but the mutual exclusion here is between the I/O thread and a single user side driver thread. The user side thread will either wait for work to complete (and needs to acquire the mutex) or will post new work (and needs the mutex as well). In either case, it has to wait for a loop tick before new work can be handled, so it doesn't matter if the lock is acquired a bit later than strictly possible (due to coarse grained locking), because it would subsequently have to wait for the event loop thread again anyway.

PR #253, opened 2020-04-06
Bump kMultiProcTimeout from 200msec->500msec.

jjlilley:
Summary:
When running some stress tests, we're encountering occasionally flakiness
in some tests like Transport/TransportMultiProcTest.UnboundIoErrors.

Particularly, some test cases use half that 200msec timeout as a signal
for failure, i.e.
  ASSERT_LT(delta.count(), kMultiProcTimeout.count() / 2);

When running myself, I see the test taking ~110ms (vs 100ms timeout)
chunk of time. Bumping up the timeout slightly will make the test more
reliable on build system hardware

Differential Revision: D20877063

### DIFF
--- a/gloo/test/multiproc_test.h
+++ b/gloo/test/multiproc_test.h
@@ -24,7 +24,7 @@ namespace gloo {
 namespace test {
 
 const int kExitWithIoException = 10;
-const auto kMultiProcTimeout = std::chrono::milliseconds(200);
+const auto kMultiProcTimeout = std::chrono::milliseconds(500);
 
 class MultiProcTest : public ::testing::Test {
  protected:


gloo/test/multiproc_test.h
```diff
@@ -24,7 +24,7 @@ namespace gloo {
 namespace test {
 
 const int kExitWithIoException = 10;
-const auto kMultiProcTimeout = std::chrono::milliseconds(200);
+const auto kMultiProcTimeout = std::chrono::milliseconds(300);
```
pritamdamania87, line 27: Does it hurt to make this a second or two to be conservative? Looks like this is the timeout for rendezvous.

jjlilley: That's a reasonable point, let me bump it up a bit more to 500msec, and see what that does to test reliability. If it creeps up too high, my feeling is that any explicit timeout-type tests are going to take longer.

In a previous situation, we had an effective SLA that timing-related unittests on build machines should be resilient to a 100ms hiccup at any time - and seemed to make things fairly reliable.  This seems to get us there, and we can revisit if it's not sufficient.

jjlilley:
Is there something that I need to do to land this? When I did "Land to fbsource" it transferred over, but the approval didn't (I've tried a couple times)
   Is there a special permission for gloo? 
   Maybe it would help to approve D20877063 as well?

PR #260, opened 2020-07-17
OpenSSL-over-TCP transport

pbelevich:
### Motivation
The purpose of this pull request is to support TLS in Gloo. This is done by creating new TLS-over-TCP transport which extends existing TCP transport with OpenSSL library.

### How to use
To migrate from TCP transport to TLS-over-TCP just replace
`::gloo::transport::tcp::CreateDevice(attr)`
with
`::gloo::transport::tcp::tls::CreateDevice(attr, pkey_file_path, cerf_file_path)`
where `pkey_file_path` is a path to Private Key file and `cerf_file_path` is a path to Certificate file


.tsan-suppressions
```diff
@@ -9,3 +9,7 @@ race:gloo::transport::tcp::Address::operator=
 # won't call into it before its destructor returns.
 # See https://github.com/facebookincubator/gloo/pull/236.
 race:gloo::transport::tcp::Pair::~Pair
+
+# https://github.com/facebookincubator/gloo/pull/210#discussion_r322192079
```
mrshenli, line 13: any reason that we need to suppress tsan errors ini these two functions?

gloo/test/allgather_test.cc
```diff
@@ -28,9 +28,9 @@ class AllgatherTest : public BaseTest,
 
 TEST_P(AllgatherTest, VarNumPointer) {
   const auto transport = std::get<0>(GetParam());
-  const auto contextSize = std::get<0>(GetParam());
```
mrshenli, line 31: was this a bug in existing Gloo code? Do you why this test didn't fail in CI?

gloo/test/base_test.h
```diff
@@ -54,19 +59,26 @@ class Barrier {
 
 enum Transport {
   TCP,
+  TCP_TLS,
```
mrshenli, line 62: do we need a `#if` guard for this as well?

gloo/test/multiproc_test.h
```diff
@@ -64,18 +65,27 @@ class MultiProcWorker {
         new ::gloo::rendezvous::FileStore(storePath));
     semaphore_ = sem_open(semaphoreName.c_str(), 0);
     GLOO_ENFORCE_NE(semaphore_, (sem_t*)nullptr, strerror(errno));
+    cerf_file_path = std::string(std::tmpnam(nullptr)); // revert!
```
mrshenli, line 68: `std::move`?

gloo/test/multiproc_test.h
```diff
@@ -64,18 +65,27 @@ class MultiProcWorker {
...
+    pkey_file_path  = std::string(std::tmpnam(nullptr)); // revert!
+    create_pkey_and_cert(pkey_file_path, cerf_file_path); // revert!
   }
 
   ~MultiProcWorker() {
     sem_close(semaphore_);
+    std::remove(pkey_file_path.c_str()); // revert!
```
mrshenli, line 75: why its default dtor does not work?

gloo/test/openssl_utils.cc
```diff
@@ -0,0 +1,94 @@
...
+#define wrap_in_unique_ptr(NAME, TYPE) \
+  wrap_in_unique_ptr_prefix(NAME, TYPE, TYPE)
+
+void create_pkey_and_cert(const std::string& pkey_file_path,
+                          const std::string& cert_file_path) {
+  wrap_in_unique_ptr(pkey, EVP_PKEY);
+
+  wrap_in_unique_ptr_prefix(bn, BIGNUM, BN);
+
+  GLOO_ENFORCE(BN_set_word(bn, RSA_F4),
```
mrshenli, line 32: can we run clang-format on code? I wonder if clang-format would give the following format:

```c++
GLOO_ENFORCE(
    BN_set_word(bn, RSA_F4),
    "BN_set_word failed");
```

gloo/test/openssl_utils.cc
```diff
@@ -0,0 +1,94 @@
...
+
+#include <openssl/pem.h>
+#include <openssl/x509.h>
+
+#include <memory>
+
+namespace gloo {
+namespace test {
+
+#define wrap_in_unique_ptr_prefix(NAME, TYPE, PREFIX)                         \
```
mrshenli, line 13: Curious, why macro name not in all capital letters? Is this the convention in Gloo code base?

gloo/test/openssl_utils.cc
```diff
@@ -0,0 +1,94 @@
...
+auto NAME = NAME##_unique_ptr.get()
+
+#define wrap_in_unique_ptr(NAME, TYPE) \
+  wrap_in_unique_ptr_prefix(NAME, TYPE, TYPE)
+
+void create_pkey_and_cert(const std::string& pkey_file_path,
+                          const std::string& cert_file_path) {
+  wrap_in_unique_ptr(pkey, EVP_PKEY);
+
+  wrap_in_unique_ptr_prefix(bn, BIGNUM, BN);
```
mrshenli, line 30: IIUC, `BIGNUM` will be used as a type to create `BIGNUM*`. Where is `BIGNUM` defined?

mrshenli, line 30: Any reason for using macro to create a variable? This is not immediately clear that this will define a `bn_unique_ptr` var.

gloo/test/transport_test.cc
```diff
@@ -188,7 +188,7 @@ TEST_P(TransportMultiProcTest, UnboundIoErrors) {
   wait();
   const auto delta = std::chrono::duration_cast<std::chrono::milliseconds>(
       std::chrono::high_resolution_clock::now() - start);
-  ASSERT_LT(delta.count(), kMultiProcTimeout.count() / 2);
+  ASSERT_LT(delta.count(), kMultiProcTimeout.count() * 2);
```
mrshenli, line 191: what's the reason for this change?

gloo/transport/tcp/tls/pair.cc
```diff
@@ -0,0 +1,386 @@
...
+ *
+ * This source code is licensed under the BSD-style license found in the
+ * LICENSE file in the root directory of this source tree.
+ */
+
+
+#include "gloo/transport/tcp/tls/pair.h"
+
+#include "gloo/common/error.h"
+#include <gloo/common/logging.h>
```
mrshenli, line 13: any reason this needs to be <>?

gloo/transport/tcp/tls/pair.cc
```diff
@@ -0,0 +1,386 @@
...
+
+    rx_.nread += rv;
+  }
+
+  // Read completed
+  const auto opcode = rx_.getOpcode();
+  switch (opcode) {
+    case Op::SEND_BUFFER:
+      // Done sending data to pinned buffer; trigger completion.
+      rx_.buf->handleRecvCompletion();
```
mrshenli, line 234: any reason TLS requires `handleRecvCompletion` to be a public API and call it here while other transports do not?

gloo/transport/tcp/tls/pair.cc
```diff
@@ -0,0 +1,386 @@
...
+  // Pair::changeState and Device::unregisterDescriptor for more info.
+  GLOO_ENFORCE_LE(state_, CONNECTED);
+
+  // Exception must not be set.
+  // If exception is set, state must advance to CLOSED state.
+  GLOO_ENFORCE(ex_ == nullptr);
+
+  if (state_ == CONNECTED) {
+    if (!is_ssl_connected_) {
+      if (!is_client_) {
```
mrshenli, line 281: compared to `tcp/pair.cc`'s implementation, looks like this `if` branch is the only part that differs. I wonder if we can we can come up with a solution that avoid code duplication.

gloo/transport/tcp/tls/pair.cc
```diff
@@ -0,0 +1,386 @@
...
+      int events;
+      while ((events = handshake())) {
+        int r;
+        struct pollfd pfd;
+        pfd.fd = fd_;
+        pfd.events = events;
+        do {
+          r = poll(&pfd, 1, 100);
+        } while (r == 0);
+        GLOO_ENFORCE(r == 1, "poll return %d error events: %d errno %d %s\n",
```
mrshenli, line 361: is it true that there is only one socket so that `r` should also equal to 1? BTW, can we add a link to poll's return value explanation in comments?

gloo/transport/tcp/tls/pair.cc
```diff
@@ -0,0 +1,386 @@
...
+    handleConnecting();
+    return;
+  }
+
+  GLOO_ENFORCE(false, "Unexpected state: ", state_);
+}
+
+void Pair::changeState(Pair::state nextState) noexcept {
+  if (nextState == CLOSED && is_ssl_connected_) {
+    SSL_shutdown(ssl_);
```
mrshenli, line 329: do we need to check return value of `SSL_shutdown` in case it is not successful? https://www.openssl.org/docs/man1.0.2/man3/SSL_shutdown.html

gloo/test/allgather_test.cc
pbelevich: Yes it was and since `std::get<0>(GetParam());` returned `0`, the `contextSize ` was `0` and the tests just didn't do anything

gloo/test/multiproc_test.h
pbelevich: sorry, I don't understand what you mean, it's deletion of `pkey_file_path` file

gloo/test/openssl_utils.cc
pbelevich: uppercased

gloo/test/openssl_utils.cc
pbelevich: in `openssl/ossl_typ.h` via `openssl/x509.h`

gloo/transport/tcp/pair.cc
```diff
@@ -234,8 +235,10 @@ void Pair::connect(const Address& peer) {
     GLOO_THROW_INVALID_OPERATION_EXCEPTION("cannot connect to self");
   }
 
+  is_client_ = rv > 0;
+
   // self_ < peer_; we are listening side.
-  if (rv < 0) {
+  if (!is_client_) {
```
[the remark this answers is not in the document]
pbelevich: I need `is_client_ ` flag in the `tls::Pair` sub class

gloo/test/multiproc_test.h
pbelevich: I don't know how `std::move` might be applicable here, because the right-hand side is already `std::string&&`, so this line calls move assignment operator

gloo/test/openssl_utils.cc
pbelevich: clang-format gave `GLOO_ENFORCE(BN_set_word(bn, RSA_F4), "BN_set_word failed");`

gloo/test/transport_test.cc
pbelevich: To fix tests, if test passes with `2*t` timeout, but doesn't with `t` timeout, it should be adjusted, shouldn't it?

gloo/transport/tcp/tls/pair.cc
pbelevich: fixed

gloo/transport/tcp/tls/pair.cc
pbelevich: The existing code based on friend classes and this prevents access from sub-classes

gloo/transport/tcp/pair.h
```diff
@@ -287,14 +290,31 @@ class Pair : public ::gloo::transport::Pair, public Handler {
   //
   // The pair mutex is expected to be held when called.
   //
-  void changeState(state nextState) noexcept;
+  virtual void changeState(state nextState) noexcept;
+
+  template<typename pred_t>
+  void waitUntil(pred_t pred,
+                       std::unique_lock<std::mutex>& lock,
```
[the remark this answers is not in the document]
pbelevich: addressed

gloo/transport/tcp/pair.h
```diff
@@ -25,6 +25,7 @@
 #include <sys/socket.h>
 #include <sys/uio.h>
 
+#include <gloo/common/error.h>
```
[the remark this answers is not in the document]
pbelevich: addressed

gloo/transport/tcp/tls/context.cc
```diff
@@ -0,0 +1,65 @@
...
+ * Copyright (c) 2020-present, Facebook, Inc.
+ * All rights reserved.
+ *
+ * This source code is licensed under the BSD-style license found in the
+ * LICENSE file in the root directory of this source tree.
+ */
+
+#include "gloo/transport/tcp/tls/context.h"
+
+#include <gloo/common/logging.h>
```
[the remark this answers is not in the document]
pbelevich: addressed

gloo/transport/tcp/tls/context.cc
```diff
@@ -0,0 +1,65 @@
...
+
+Context::~Context() {
+  SSL_CTX_free(ssl_ctx_);
+}
+
+std::unique_ptr<transport::Pair>& Context::createPair(int rank) {
+  pairs_[rank] = std::unique_ptr<transport::Pair>(
+      new tls::Pair(this,
+          dynamic_cast<tls::Device*>(device_.get()),
+          rank, getTimeout())
```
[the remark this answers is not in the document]
pbelevich: addressed

gloo/transport/tcp/tls/pair.cc
```diff
@@ -0,0 +1,386 @@
...
+namespace tcp {
+namespace tls {
+
+Pair::Pair(
+    Context* context,
+    Device* device,
+    int rank,
+    std::chrono::milliseconds timeout)
+    : ::gloo::transport::tcp::Pair(context, device, rank, timeout),
+      ssl_(nullptr), is_ssl_connected_(false) {
```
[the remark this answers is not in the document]
pbelevich: done

.tsan-suppressions
pbelevich: the reason is [the same as it was for writev/recv](https://github.com/facebookincubator/gloo/pull/210#discussion_r322192079) but now it is for SSL_read/SSL_write

gloo/test/base_test.h
pbelevich: instead I decided to remove #if guards from this file to make it consistent with TCP/UV transports

gloo/test/multiproc_test.h
```diff
@@ -18,7 +18,8 @@
 #include "gloo/common/logging.h"
 #include "gloo/rendezvous/context.h"
 #include "gloo/rendezvous/file_store.h"
-#include "gloo/transport/tcp/device.h"
+#include "gloo/transport/tcp/tls/device.h" // revert!
```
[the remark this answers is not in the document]
pbelevich: this is a temporary testing TLS_TCP instead of TCP to make sure that it's ok, I will come up with permanent solution soon

gloo/transport/tcp/tls/context.cc
```diff
@@ -0,0 +1,65 @@
...
+          device->getCertificateFilePath().c_str()
+          )) {}
+
+SSL_CTX* Context::create_SSL_CTX(const char* pkey, const char* cert) {
+  static std::once_flag ssl_ctx_init_;
+  std::call_once(ssl_ctx_init_, [] {
+    SSL_load_error_strings();
+    SSL_library_init();
+  });
+  SSL_CTX* ssl_ctx = SSL_CTX_new(SSLv23_method());
```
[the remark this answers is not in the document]
pbelevich: addressed

mrshenli:
Hey @pbelevich, let me know when this is ready for a second round of review. thanks.

gloo/test/openssl_utils.cc
pbelevich: this file was totally rewritten, the comment is obsolete

gloo/transport/tcp/tls/pair.cc
```diff
@@ -0,0 +1,386 @@
...
+}
+
+Pair::~Pair() {}
+
+inline SSL_CTX *Pair::get_ssl_ctx() {
+  return dynamic_cast<Context*>(context_)->ssl_ctx_;
```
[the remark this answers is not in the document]
pbelevich: addressed. moved to c-tor

gloo/transport/tcp/tls/pair.cc
pbelevich: refactored, `handleReadWrite()` was extracted

gloo/transport/tcp/tls/pair.cc
```diff
@@ -0,0 +1,375 @@
...
+  GLOO_ENFORCE(!is_ssl_connected_);
+
+  if (ssl_ == nullptr) {
+    GLOO_ENFORCE(get_ssl_ctx() != nullptr);
+    ssl_ = SSL_new(get_ssl_ctx());
+    GLOO_ENFORCE(ssl_ != nullptr, "SSL_new failed");
+    int r = SSL_set_fd(ssl_, fd_);
+    GLOO_ENFORCE(r, "SSL_set_fd failed");
+    if (is_client_) {
+      SSL_set_connect_state(ssl_);
```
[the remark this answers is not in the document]
pbelevich: addressed

gloo/test/openssl_utils.cc
```diff
@@ -0,0 +1,89 @@
...
+                                                      TYPE##_deleter);         \
+    GLOO_ENFORCE(x.get(), #PREFIX "_new failed");                              \
+    return x;                                                                  \
+  }();                                                                         \
+  auto NAME = NAME##_unique_ptr.get()
+
+#define WRAP_IN_UNIQUE_PTR(NAME, TYPE)                                         \
+  WRAP_IN_UNIQUE_PTR_PREFIX(NAME, TYPE, TYPE)
+
+void create_pkey_and_cert(const std::string &pkey_file_path,
```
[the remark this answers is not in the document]
pbelevich: addressed, rewriten

gloo/transport/tcp/tls/pair.cc
pbelevich: It can return negative value according to the doc

gloo/transport/tcp/tls/context.cc
```diff
@@ -0,0 +1,79 @@
...
+               [](::SSL_CTX *x) { ::SSL_CTX_free(x); }) {}
+
+SSL_CTX *Context::create_ssl_ctx(const char *pkey, const char *cert,
+                                 const char *ca_file, const char *ca_path) {
+  static std::once_flag ssl_ctx_init_;
+  std::call_once(ssl_ctx_init_, [] {
+    SSL_load_error_strings();
+    SSL_library_init();
+  });
+  SSL_CTX *ssl_ctx = SSL_CTX_new(SSLv23_method());
```
pbelevich, line 45: change to `TLSv1_2_method`

pbelevich: changed

.circleci/config.yml
```diff
@@ -62,8 +62,6 @@ jobs:
             cd build
             cmake ../ \
               -DCMAKE_VERBOSE_MAKEFILE=ON \
-              -DUSE_IBVERBS=ON \
-              -DUSE_LIBUV=ON \
```
mrshenli, line 66: Why moving this to individual builds and distinguish `*` with `*-all-transports`?

.circleci/config.yml
```diff
@@ -78,14 +76,20 @@ jobs:
 
             # Exclude multi-process transport tests because
             # they appear to have some crashing problem...
-            EXCLUDE="Transport/TransportMultiProcTest*:$EXCLUDE"
+            # EXCLUDE="Transport/TransportMultiProcTest*:$EXCLUDE"
```
mrshenli, line 79: debug vestige?

.circleci/config.yml
```diff
@@ -78,14 +76,20 @@ jobs:
...
 
             # Exclude memory leak test for tsan job (false positive).
-            if [ "$CIRCLE_JOB" == "tsan" ]; then
```
mrshenli, line 84: is it true that this is not working before this change as it is not doing regex match?

gloo/benchmark/options.cc
```diff
@@ -150,6 +150,10 @@ struct options parseOptions(int argc, char** argv) {
       {"ib-port", required_argument, nullptr, 0x100f},
       {"tcp-device", required_argument, nullptr, 0x1010},
       {"base", required_argument, nullptr, 0x1011},
+      {"tls-pkey", required_argument, nullptr, 0x2001},
```
mrshenli, line 153: curious, what is the `nullptr` for?

gloo/test/CMakeLists.txt
```diff
@@ -5,13 +5,15 @@ set(GLOO_TEST_SRCS
...
   "${CMAKE_CURRENT_SOURCE_DIR}/barrier_test.cc"
+  "${CMAKE_CURRENT_SOURCE_DIR}/base_test.cc"
   "${CMAKE_CURRENT_SOURCE_DIR}/broadcast_test.cc"
   "${CMAKE_CURRENT_SOURCE_DIR}/gather_test.cc"
   "${CMAKE_CURRENT_SOURCE_DIR}/gatherv_test.cc"
   "${CMAKE_CURRENT_SOURCE_DIR}/main.cc"
   "${CMAKE_CURRENT_SOURCE_DIR}/memory_test.cc"
   "${CMAKE_CURRENT_SOURCE_DIR}/reduce_test.cc"
   "${CMAKE_CURRENT_SOURCE_DIR}/send_recv_test.cc"
+  "${CMAKE_CURRENT_SOURCE_DIR}/openssl_utils.cc"
```
mrshenli, line 16: sort the order?

gloo/test/multiproc_test.h
```diff
@@ -15,23 +15,34 @@
...
 
 namespace gloo {
 namespace test {
 
 const int kExitWithIoException = 10;
 const auto kMultiProcTimeout = std::chrono::milliseconds(500);
 
+const std::vector<Transport> kTransportsForMultiProcTest {
+#if GLOO_HAVE_TRANSPORT_TCP_TLS
+    Transport::TCP_TLS,
```
mrshenli, line 34: Would I be correct if I assume adding this here will make all existing tests run in both TCP and TCP_TLS modes?

gloo/test/multiproc_test.h
```diff
@@ -15,23 +15,34 @@
...
+    Transport::TCP,
+};
+
 class MultiProcTest : public ::testing::Test {
  protected:
   void SetUp() override;
   void TearDown() override;
 
-  void spawnAsync(int size, std::function<void(std::shared_ptr<Context>)> fn);
+  void spawnAsync(Transport transport, int size,
```
mrshenli, line 44: is this what clang-format gives? shouldn't it be either one arg per-line or all args in one line?

gloo/test/openssl_utils.cc
```diff
@@ -0,0 +1,43 @@
...
+#include <cstdlib>
+#include <gloo/common/logging.h>
+
+namespace gloo {
+namespace test {
+
+std::string ca_key_file;
+std::string ca_file;
+std::string pkey_file;
+std::string csr_file;
```
mrshenli, line 13: didn't see this one in benchmark options, what is this string?

gloo/test/openssl_utils.h
```diff
@@ -0,0 +1,19 @@
...
+#include <string>
+
+namespace gloo {
+namespace test {
+
+extern std::string ca_key_file;
+extern std::string ca_file;
+extern std::string pkey_file;
+extern std::string csr_file;
+extern std::string cert_file;
```
mrshenli, line 15: names here are inconsistent with the ones in benchmark options, any reason for this?

gloo/transport/tcp/pair.cc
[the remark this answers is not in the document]
mrshenli: I mean the condition in this if clause is different than before (`<` vs `<=`). Is this expected?

gloo/transport/tcp/pair.h
```diff
@@ -251,14 +257,18 @@ class Pair : public ::gloo::transport::Pair, public Handler {
...
+
+  void readComplete(NonOwningPtr<UnboundBuffer> &buf);
 
   // Helper function that is called from the `read` function.
   void handleRemotePendingSend(const Op& op);
 
   // Helper function that is called from the `read` function.
   void handleRemotePendingRecv(const Op& op);
 
+  virtual void handleReadWrite(int events);
```
mrshenli, line 270: add comments for it?

gloo/transport/tcp/tls/pair.cc
```diff
@@ -0,0 +1,284 @@
...
+          if (errno == EPIPE) {
+            if (!sync_) {
+              return false;
+            }
+          }
+          if (errno == ECONNRESET) {
+            if (!sync_) {
+              return false; // Exit if Connection reset by peer on read
+            }
+          }
```
mrshenli, line 189: any reason these are add as two different branches? will `errno == EPIPE  || errno == ECONNRESET` be sufficient?

gloo/transport/tcp/tls/pair.cc
```diff
@@ -0,0 +1,284 @@
...
+        if (err == SSL_ERROR_ZERO_RETURN) {
+          if (errno == ECONNRESET) {
+            if (!sync_) {
+              return false;
+            }
+          }
+        }
+
+        if (err == SSL_ERROR_SYSCALL) {
+          fatal_error_occurred_ = true;
```
mrshenli, line 179: what's the definition of fatal error here? why unexpected error is not fatal?

mrshenli:
cc @agolynski any thoughts on how the API should looks like in c10d?

.circleci/config.yml
pbelevich: To ensure that the project can be built with and without any transport(that all preprocessor if-branches are valid)

.circleci/config.yml
pbelevich: Looks like I found and fixed the existing issues with TCP transport and TransportMultiProcTest, but I wanted to preserve this lines just in case for sometime...

.circleci/config.yml
pbelevich: yes, it is, the original job has name "tsan"

gloo/benchmark/options.cc
pbelevich: > If the field 'flag' is not NULL, it points to a variable that is set
>    to the value given in the field 'val' when the option is found, but
>    left unchanged if the option is not found.
> 
>    To have a long-named option do something other than set an 'int' to
>    a compiled-in constant, such as set a value from 'optarg', set the
>    option's 'flag' field to zero and its 'val' field to a nonzero
>    value (the equivalent single-letter option character, if there is
>    one).  For long options that have a zero 'flag' field, 'getopt'
>    returns the contents of the 'val' field.

gloo/test/CMakeLists.txt
pbelevich: addressed

gloo/test/multiproc_test.h
pbelevich: yes, except not all, but only Transport tests

gloo/transport/tcp/tls/pair.cc
pbelevich: > Note that SSL_shutdown() must not be called if a previous fatal error has occurred on a connection i.e. if SSL_get_error() has returned SSL_ERROR_SYSCALL or SSL_ERROR_SSL.

gloo/test/multiproc_test.h
pbelevich: my clang-format gives this

gloo/test/openssl_utils.cc
pbelevich: deleted

gloo/transport/tcp/pair.h
pbelevich: addressed

gloo/transport/tcp/tls/pair.cc
pbelevich: addressed

gloo/test/openssl_utils.h
pbelevich: addressed

gloo/transport/tcp/tls/pair.cc
pbelevich: After some investigations I think that the best what I can do is just call `SSL_shutdown(ssl_)` twice

gloo/transport/tcp/tls/pair.cc
pbelevich: I added a comment

gloo/transport/tcp/tls/pair.cc
pbelevich: I added this field not to do `SSL_shutdown` in case of `SSL_ERROR_SYSCALL`

.circleci/config.yml
mrshenli: Can we add a comment here to explain why we leave this here?

PR #261, opened 2020-07-21
HIP library renamed with move to hipclang

jeffdaily:


jeffdaily:
This PR is needed to fully resolve https://github.com/pytorch/pytorch/pull/41813.

cmake/Hip.cmake
```diff
@@ -112,9 +112,16 @@ find_package(HIP 1.0)
 IF(HIP_FOUND)
   set(HAVE_HIP TRUE)
 
+  if(HIP_COMPILER STREQUAL clang)
```
xw285cornell, line 115: Will it be a good idea to define HIP_COMPILER in gloo also, in case gloo is independently compiled?

jeffdaily: HIP_COMPILER is set by `find_package(HIP 1.0)`.  The FindHIP.cmake does this.

PR #264, opened 2020-07-31
Enable Gloo working on windows platform, only support uv as transport

gunandrose4u:
## What have done

1. Add MSVC to detect whether compiling for windows platform
2. Force to use libuv as transport way
3. Exclude src directories, nccl, transport/tcp, transport/ibverbs
4. Disable compile option like USE_REDIS_DEFAULT, USE_IBVERBS_DEFAULT etc.
5. Add libuv library as submodule
6. Add a new example to show all_reduce function and libuv as transport since currently example is leverage tcp as transport
7. Port unsupported gloo source code to MSVC
8. Port unsupported gloo/cuda* source code to MSVC
9. Port test case (NOT include cuda part due to build issue) related to all_reduce and broadcast function to MSVC


.gitmodules
```diff
@@ -1,3 +1,6 @@
 [submodule "third-party/googletest"]
 	path = third-party/googletest
 	url = https://github.com/google/googletest.git
+[submodule "third_party/libuv"]
```
pbelevich, line 4: I understand that it will be the default transport implementation on Windows, but it's already the default transport implementation on macOS and we didn't make it github submodule, so why do we need to make it github submodule now? do we really need to make it github submodule? maybe it makes sense to expect it as an installed dependency? cc @mrshenli

gunandrose4u: Since in PyTorch libuv is a sub module of tensorpipe, and for windows, just enable gloo, so I made it as a sub module of gloo.  My thinking is if both tensorpipe and gloo enabled when building PyTorch, there is only one libuv target can be created, so libuv will not be built twice. I've not tried to install libuv on windows, let me try it. 
Do you mean that on macOS by default gloo will use libuv?
@mrshenli ,

gunandrose4u:
> @gunandrose4u, thank you for working on this feature and sorry for the late review.
> 
> 1. First of all please sign CLA
> 2. Please add windows build to `.circleci/config.yml`, circleci has a good [documentaion](https://circleci.com/docs/2.0/configuration-reference/) and we have a plenty of examples in [pytorch](https://github.com/pytorch/pytorch/blob/master/.circleci/config.yml) or [vision](https://github.com/pytorch/vision/blob/master/.circleci/config.yml)

I added a Windows build for it. And will enable test later.

.gitmodules
mrshenli: > Do you mean that on macOS by default gloo will use libuv?

Yes, on OSX, Gloo will use libuv, and we currenct expect OSX users to install pkg-config and libuv through conda. 

Just checked TensorPipe did include libuv as a submodule, and recall there was some issue with that during build. 

cc @beauby @malfet Since both TensorPipe and Gloo will use libuv, what's the recommended way here? should we pull libuv as a submodule into PyTorch, or should we keep them separate?

gunandrose4u: @beauby @malfet I will work on find_package on Windows for Libuv, will update the it in this PR.

pbelevich:
> I just tried to enable transport and multiproc test, I found it base on tcp transport. I will make a version with uv as transport.

@gunandrose4u please take a look at gloo/test/transport_test.cc in https://github.com/facebookincubator/gloo/pull/260

Let's have something like this:
```
const std::vector<Transport> kTransportsForMultiProcTest {
#if GLOO_HAVE_TRANSPORT_TCP_TLS
    Transport::TCP_TLS,
#endif
#if GLOO_HAVE_TRANSPORT_TCP
    Transport::TCP_TLS,
#endif
#if GLOO_HAVE_TRANSPORT_UV
    Transport::UV,
#endif
};
```

WDYT?

gunandrose4u:
> > I just tried to enable transport and multiproc test, I found it base on tcp transport. I will make a version with uv as transport.
> 
> @gunandrose4u please take a look at gloo/test/transport_test.cc in #260
> 
> Let's have something like this:
> 
> ```
> const std::vector<Transport> kTransportsForMultiProcTest {
> #if GLOO_HAVE_TRANSPORT_TCP_TLS
>     Transport::TCP_TLS,
> #endif
> #if GLOO_HAVE_TRANSPORT_TCP
>     Transport::TCP_TLS,
> #endif
> #if GLOO_HAVE_TRANSPORT_UV
>     Transport::UV,
> #endif
> };
> ```
> 
> WDYT?

When I tried to enable TransportMultiProcTest, I found another thing is that it depends on fork API to start child processes to run worker. But on Windows platform, there is no API like fork.  One way is to implement a fork API, but it is not easy and will be tricky, not stable. Other way is to leverage CreateProcess API of Windows, in that way, need implement a stand alone multiproctest.exe which will contain logic in spawnAsync method, and also some windows message handler to simulate the SIGSTOP and SIGKILL behavior. Due to it may take lot of efforts to enable TransportMultiProcTest on Windows and in test_c10d.py in PyTorch there are cases to cover multiple processes scenario. I will enable TransportMultiProcTest in another PR.

.gitmodules
beauby: @gunandrose4u: For Windows I suggest using [`find_library()`](https://cmake.org/cmake/help/latest/command/find_library.html), and relying on the user to provide `libuv_ROOT` to CMake. Similarly, you could use `find_file()` for the headers.

gunandrose4u: Sure, let me figure out and switch to find_library().

gloo/examples/example_allreduce.cc
```diff
@@ -0,0 +1,187 @@
...
+  try{
+    context->connectFullMesh(prefixStore, dev);
+    std::cout << "connectFullMesh" << std::endl;
+    // All connections are now established. We can now initialize some
+    // test data, instantiate the collective algorithm, and run it.
+    std::array<int, 4> data;
+    std::cout << "Input: " << std::endl;
+    for (int i = 0; i < data.size(); i++) {
+      data[i] = i;
+      std::cout << "data[" << i << "] = " << data[i] << std::endl;
```
mrshenli, line 126: these are debug vestige?

gloo/examples/example_allreduce.cc
```diff
@@ -0,0 +1,187 @@
...
+    // across local buffers as well. If you have a single buffer only,
+    // you must pass a std::vector with a single pointer.
+    std::vector<int*> ptrs;
+    ptrs.push_back(&data[0]);
+
+    // The number of elements at the specified pointer.
+    int count = data.size();
+    
+    // Instantiate the collective algorithm.
+    /*
```
mrshenli, line 140: ditto?

gloo/examples/example_allreduce.cc
gunandrose4u: yes, for debugging purpose, will remove them all.

gloo/allreduce_bcube.h
```diff
@@ -16,6 +16,10 @@
 #include <iostream>
 #include <unordered_map>
 
+#ifdef _WIN32
```
Orvid, line 19: This doesn't need to be conditionally guarded, it's required for correctness. Other platforms just ended up including it as part of other STL headers.

gloo/allreduce_ring_chunked.h
```diff
@@ -32,7 +32,11 @@ class AllreduceRingChunked : public Algorithm {
     // Use chunks of no less than 1024 bytes (256 * sizeof(float))
     constexpr unsigned long minSize = 256;
```
Orvid, line 33: Just make this an `unsigned long long`

gloo/common/memory.h
```diff
@@ -140,7 +144,12 @@ class ShareableNonOwningPtr final {
     ptr_.reset();
     // Wait for all shared_ptr's to T have been released
     while (!weak.expired()) {
+#ifdef _WIN32
+      //Due to here sleep time is 0, does not want implement a usleep on windows
+      Sleep(0);
+#else
       usleep(0);
+#endif
     }
```
Orvid, line 153: `#include <thread>` above, then just use `std::this_thread::yield()` for a cross-platform version with the same effect.

gloo/cuda_allreduce_ring_chunked.cc
```diff
@@ -71,7 +71,11 @@ CudaAllreduceRingChunked<T, W>::CudaAllreduceRingChunked(
   // (256 * sizeof(float)).
   constexpr unsigned long minSize = 256;
```
Orvid, line 72: Same as above, just make this explicitly an `unsigned long long`. `long` is 64-bit on linux, but 32-bit on windows, so it's far better to make this explicit. (I would recommend searching through gloo for any references to bare `long` and fixing them, otherwise you're likely to run into issues later)

gloo/test/allreduce_test.cc
```diff
@@ -26,8 +26,13 @@ namespace {
 
 // RAII handle for aligned buffer
 template <typename T>
+#ifdef _WIN32
```
Orvid, line 29: What issue was the aligned allocator causing?

gloo/test/main.cc
```diff
@@ -12,7 +12,9 @@
 namespace {
 struct Initializer {
   Initializer() {
+#ifndef _WIN32
     signal(SIGPIPE, SIG_IGN);
+#endif
```
Orvid, line 17: Widen this `#ifndef` to cover the entire `Initializer` struct and use, there's no point in having it call the empty function on startup.

gloo/transport/unbound_buffer.h
```diff
@@ -98,7 +98,7 @@ class UnboundBuffer {
   // If the byte count argument is not specified, it will default the
   // number of bytes to be equal to the number of bytes remaining in
   // the buffer w.r.t. the offset.
-  static constexpr auto kUnspecifiedByteCount = std::numeric_limits<size_t>::max();
+  static constexpr auto kUnspecifiedByteCount = (std::numeric_limits<size_t>::max)();
```
Orvid, line 101: `#define NOMINMAX 1` wherever `Windows.h` is being included, or preferably at the build system level. No reason to have to fix it at every call site.

gloo/transport/uv/libuv.h
```diff
@@ -32,6 +32,10 @@
 #include <utility>
 #include <vector>
 
+#ifdef _WIN32
```
Orvid, line 35: If it's a standard library header, just include it unconditionally. No reason to guard it.

gloo/transport/uv/libuv.h
```diff
@@ -280,13 +284,21 @@ class Resource : public Emitter<T>, public std::enable_shared_from_this<T> {
 
   template <typename R>
   const R* get() const noexcept {
+  #ifdef _WIN32
+    static_assert(!std::is_same<R, U>::value, "!");
+  #else
     static_assert(not std::is_same<R, U>::value, "!");
+  #endif
```
Orvid, line 291: Use the `!` version for all platforms, here and in all locations below, no need to break it out like this. They are lexically identical, the text form is just a legacy of C that really needs to go away.

gloo/types.cc
```diff
@@ -9,6 +9,9 @@
 #include "gloo/types.h"
 
 #include <stdexcept>
+#ifdef _WIN32
+#include <string>
+#endif
```
Orvid, line 14: Include this unconditionally.

gloo/types.h
```diff
@@ -84,7 +89,11 @@ struct float16;
 float16 cpu_float2half_rn(float f);
 float cpu_half2float(float16 h);
 
+#ifndef _WIN32
 struct __attribute__((__aligned__(2))) float16 {
+#else
+__declspec(align(2)) struct float16 {
+#endif
```
Orvid, line 96: `alignas(2)`

gloo/types.h
```diff
@@ -104,6 +113,13 @@ struct __attribute__((__aligned__(2))) float16 {
     x = res.x;
   }
 
+#ifdef _WIN64
```
Orvid, line 116: Make the `unsigned long` above into `unsigned long long` to match semantics with Linux.

mrshenli:
Hey @Orvid, could you please help check if the latest version looks OK to you? Thanks!

gloo/test/memory_test.cc
```diff
@@ -11,18 +11,33 @@
...
   size_t size;
   size_t resident;
   f >> size >> resident;
+#ifdef _WIN32
+  SYSTEM_INFO si;
+  GetSystemInfo(&si);
+  return (si.dwPageSize * resident);
+#else
   return (getpagesize() * resident);
+#endif
```
Orvid, line 40: Don't need `<process.h>`, and the current code won't work for getting the windows resident set size. On windows you can replace the entire function with:
```
PROCESS_MEMORY_COUNTERS counters{};
GetProcessMemoryInfo(GetCurrentProcess(), &counters, sizeof(counters));
return counters.WorkingSetSize;
```

gloo/test/allreduce_test.cc
Orvid: Probably fine for now, but in the future `aligned_allocator` can be implemented using `_aligned_malloc` and `_aligned_free` on Windows, since it doesn't have `posix_memalign`.

Orvid (commented):
A couple of things left to cleanup a few more of the unnecessary `#ifdef _WIN32`s, but otherwise this looks good :)

PR #267, opened 2020-09-17
Fix build failure together with PyTorch Windows CI

gunandrose4u:
Avoid to set CMAKE_CXX_COMPILER_LAUNCHER and CMAKE_C_COMPILER_LAUNCHER to sccache which will lead a wrong compile command generated, as follow. And also fix creating device by host name error on Windows platform.

```
F:\pytorchtmp\build_ci\bin\sccache.exe F:\pytorchtmp\build_ci\bin\sccache-cl.exe   /TP -DNOMINMAX=1 -DTH_BLAS_MKL -D_OPENMP_NOFORCE_MANIFEST -I..\cmake\..\third_party\benchmark\include -I..\third_party\gloo -Ithird_party\gloo -I..\cmake\..\third_party\googletest\googlemock\include -I..\cmake\..\third_party\googletest\googletest\include -I..\third_party\protobuf\src -Iwin_tmp\mkl\include -I..\third_party -I..\cmake\..\third_party\eigen -IF:\Users\jozh\anaconda3\include -IF:\Users\jozh\anaconda3\Lib\site-packages\numpy\core\include -I..\cmake\..\third_party\pybind11\include -IF:\Temp\libuv_install\include\uv.h\.. /DWIN32 /D_WINDOWS /GR /EHsc /w /bigobj -openmp:experimental /MD /O2 /Ob2 /DNDEBUG /w /bigobj -std:c++14 /showIncludes /Fothird_party\gloo\gloo\CMakeFiles\gloo.dir\algorithm.cc.obj /Fdthird_party\gloo\gloo\CMakeFiles\gloo.dir\gloo.pdb /FS -c ..\third_party\gloo\gloo\algorithm.cc
```


gunandrose4u:
@mrshenli, please help to review this PR. Thank you! It is for fix PyTorch Windows CI failure.

gloo/transport/uv/device.cc
```diff
@@ -108,6 +108,15 @@ static bool lookupAddrForIface(struct attr* attr) {
 }
 
 static void lookupAddrForHostname(struct attr& attr) {
+#ifdef _WIN32
+  WSADATA wsa_data;
```
mrshenli, line 112: Since the rest of the team have limited experiences on Windows, shall we add some comments/pointers to explain this?

cmake/Dependencies.cmake
```diff
@@ -47,23 +47,37 @@ if(USE_LIBUV)
...
+        PATH_SUFFIXES lib
         REQUIRED
         NO_DEFAULT_PATH)
-      if(NOT EXISTS ${libuv_LIBRARY})
-        message(FATAL_ERROR "Unable to find static libuv library in " ${libuv_ROOT})
+      if(EXISTS ${libuv_LIBRARY})
+        set(libuv_HEADER_SEARCH_PATH ${libuv_LIBRARY}/../../include)
+        set(libuv_DLL_PATH ${libuv_LIBRARY}/../../bin/uv.dll)
+      else()
+        find_library(
```
mrshenli, line 57: curious, there is already a `find_library` before the `if` clause, why do we need to find it again here? Is it because the files sometimes live in `lib` and sometimes live in `lib/release` and `lib/debug?

cc @malfet, does this look OK to you?

CMakeLists.txt
```diff
@@ -60,15 +60,17 @@ if(NOT CMAKE_BUILD_TYPE)
...
-find_program(SCCACHE_EXECUTABLE sccache)
-mark_as_advanced(SCCACHE_EXECUTABLE)
-if(SCCACHE_EXECUTABLE)
-  foreach(LANG CXX C)
-    if(NOT DEFINED CMAKE_${LANG}_COMPILER_LAUNCHER AND NOT CMAKE_${LANG}_COMPILER MATCHES ".*/s?ccache$")
-      message(STATUS "Enabling sccache for ${LANG}")
-      set(CMAKE_${LANG}_COMPILER_LAUNCHER ${SCCACHE_EXECUTABLE} CACHE STRING "")
-    endif()
-  endforeach()
+if(NOT MSVC)
```
mrshenli, line 63: question, looks like if it is MSVC, we are not going to configure sccache? Or is this configured somewhere else?

gloo/transport/uv/device.cc
gunandrose4u: sure, will do.

cmake/Dependencies.cmake
gunandrose4u: For libuv installed from conda, uv and uv_a libs are under lib folder, but if we build Libuv project and install on WIndows, uv and uv_a libs will under folder named lib\release or lib\debug, depends on the build config. Since when building gloo in PyTorch, need conda install libuv. But in Windows CI of Gloo, I follow the style of other CI, build the libuv from source code and install it. That is why I made it as this. If it is OK that Windows CI of Gloo works with installing libuv from conda source. That would be simple here. @malfet, how do you think?

CMakeLists.txt
gunandrose4u: From the build log, I think it is configed in build env for PyTorch. Gloo will leverage sccache-cl.exe to compile, just as other submodules in PyTorch.

gloo/transport/uv/device.cc
malfet: Good old WSAStartup 🎺 
Don't we want/need to guard initialization block with `std::call_once`?
And it looks like the same is needed for tcp transport?

CMakeLists.txt
malfet: Looks like all that is needed here is to add another  `CMAKE_${LANG}_COMPILER` pattern for Windows, wouldn't it?

gloo/transport/uv/device.cc
gunandrose4u: I do not enable tcp tranport on Windows, so currently we do not need do the same to tcp transport. Base on MSDN doc, we can call WSAStartup multiple times, it only loads DLL for first time call, for other calls, just increment a ref count. But I still guard it with std::call_once.

cmake/Dependencies.cmake
gunandrose4u: resolved

CMakeLists.txt
gunandrose4u: Yes, you are right. For windows it is a sccache-cl.exe. Resolved it.

mrshenli (approved):

PR #272, opened 2021-01-12
Prevent potential overflow issue in tcp/pair.cc

sofong5:
Summary: Prevent overflow by clamping rel_time to be maximum of (std::chrono::milliseconds::max - std::chrono::steady_clock::now())

Differential Revision: D25885663


osalpekar (approved):
Looks good!

PR #297, opened 2021-02-26
Fix compile error for rocm

wenwu-glagle:
Fix compile error for rocm:

1. add python in command line
2. print error code when "Failed to get hipify files list!" error happened.


Madouura:
This is still a problem.

PR #302, opened 2021-03-22
Dynamic loading of OpenSSL

pbelevich:
* **#302 Dynamic loading of OpenSSL**

Differential Revision: [D27781296](https://our.internmc.facebook.com/intern/diff/D27781296)


malfet (approved):

CMakeLists.txt
```diff
@@ -26,15 +26,17 @@ set(USE_IBVERBS_DEFAULT OFF)
...
 
 # Options
 option(USE_REDIS "Support using Redis for rendezvous" ${USE_REDIS_DEFAULT})
 option(USE_IBVERBS "Support ibverbs transport" ${USE_IBVERBS_DEFAULT})
 option(USE_NCCL "Support using NCCL for local collectives" ${USE_NCCL_DEFAULT})
 option(USE_RCCL "Support using RCCL for local collectives" ${USE_RCCL_DEFAULT})
 option(USE_LIBUV "Build libuv transport" ${USE_LIBUV_DEFAULT})
-option(USE_TCP_OPENSSL "Build tcp+openssl transport" ${USE_TCP_OPENSSL_DEFAULT})
+option(USE_TCP_OPENSSL_LINK "Build tcp+openssl transport" ${USE_TCP_OPENSSL_LINK_DEFAULT})
+option(USE_TCP_OPENSSL_LOAD "Build tcp+openssl transport" ${USE_TCP_OPENSSL_LOAD_DEFAULT})
```
malfet, line 39: Nits: Can you change message to "Enable SSL TLS support"  to one and "Enable SSL/TLS dynamic binding to another". 
Also, if they are mutually exclusive, can you check for that?

gloo/transport/tcp/tls/openssl/dynamic_library.cc
```diff
@@ -0,0 +1,35 @@
...
+  }
+}
+
+void *DynamicLibrary::sym(const char *name) {
+  if (handle == nullptr) {
+    throw std::runtime_error(error);
+  }
+  void* res = dlsym(handle, name);
+  if (res == nullptr) {
+    throw std::runtime_error("Can't find " + std::string(name) + " in " + lib_name + ":" + dlerror());
```
malfet, line 26: This will crash as `lib_name` is always NULL, isn't it?

gloo/transport/tcp/tls/openssl/dynamic_library.h
```diff
@@ -0,0 +1,16 @@
...
+  void operator=(const DynamicLibrary&) = delete;
+
+  DynamicLibrary(const char* name, const char* alt_name);
+
+  void* sym(const char* name);
+
+  ~DynamicLibrary();
+
+private:
+  const char* lib_name;
```
malfet, line 13: Do not store raw string array pointers, since its lifetime is not defined
```suggestion
  const std::string lib_name;
```

gloo/transport/tcp/tls/openssl/dynamic_library.h
```diff
@@ -0,0 +1,16 @@
...
+  void* handle = nullptr;
+  const char* error = nullptr;
```
malfet, line 15: Same as above: if you want to preserve an error message, explicitly copy it in to `std::string`

gloo/transport/tcp/tls/openssl/dynamic_library.cc
```diff
@@ -0,0 +1,35 @@
...
+
+#include <stdexcept>
+#include <dlfcn.h>
+
+DynamicLibrary::DynamicLibrary(const char *name, const char *alt_name) : lib_name(name) {
+  handle = dlopen(name, RTLD_LOCAL | RTLD_NOW);
+  if (!handle) {
+    if (alt_name == nullptr) {
+      error = dlerror();
+    } else {
```
malfet, line 11: Few comments here:
 - Feel free to get rid of `alt_name` as it is needed only for specific usecases in PyTorch
 - It's OK to throw runtime error from the constructor
 - If you want to preserve error message, please copy it into std::string, as message returned by dlerror() might get clobbered by subsequent `dlopen` call.

gloo/transport/tcp/tls/openssl/openssl.cc
```diff
@@ -0,0 +1,111 @@
+#include <openssl/err.h>
+#include <openssl/ssl.h>
+
+#include "dynamic_library.h"
+
+namespace {
+DynamicLibrary libssl("libssl.so", "libssl.so.1.1");
```
malfet, line 7: Please avoid global constructors, as they are called at program/library load time.
Instead static variables inside functions which are constructed during 1st call to the function.
```suggestion
DynamicLibrary& get_libssl() {
  static DynamicLibrary libssl("libssl.so", "libssl.so.1.1");
  return libssl;
}

```

gloo/transport/tcp/tls/openssl/openssl.cc
```diff
@@ -0,0 +1,111 @@
...
+} // namespace
+
+
+#define CALL_SYM(NAME, ...) \
+static auto fn = reinterpret_cast<decltype(&NAME)>(libssl.sym(__func__)); \
+return fn(__VA_ARGS__)
+
+void ERR_print_errors_cb(int (*cb) (const char *, size_t, void *), void *u) {
+  CALL_SYM(ERR_print_errors_cb, cb, u);
```
malfet, line 16: To avoid symbols clashes with the actual libssl, can those function be defined in their own namespace? Say `_glootls::`?

pbelevich: addressed

gloo/transport/tcp/tls/openssl/openssl.cc
pbelevich: addressed

gloo/transport/tcp/tls/openssl/dynamic_library.cc
pbelevich: addressed

gloo/transport/tcp/tls/openssl/dynamic_library.h
pbelevich: addressed

gloo/transport/tcp/tls/openssl/dynamic_library.h
pbelevich: addressed

CMakeLists.txt
pbelevich: addressed

PR #309, opened 2021-05-21
allow reduce and allreduce to accept size 0 input

agolynski:
Summary:
Context: https://github.com/pytorch/pytorch/issues/58467

We'd like GLOO to be conform with other backends and therefore it should accept 0 size input for reduce/allreduce.

Differential Revision: D28608970

### DIFF
--- a/gloo/allreduce.cc
+++ b/gloo/allreduce.cc
@@ -95,6 +95,10 @@ BroadcastRangeFunction genLocalBroadcastFunction(const BufferVector& out) {
 }
 
 void allreduce(const detail::AllreduceOptionsImpl& opts) {
+  if (opts.elements == 0) {
+    return;
+  }
+
   const auto& context = opts.context;
   const std::vector<std::unique_ptr<transport::UnboundBuffer>>& in = opts.in;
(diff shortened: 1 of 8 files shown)


gloo/allreduce.cc
```diff
@@ -95,6 +95,10 @@ BroadcastRangeFunction genLocalBroadcastFunction(const BufferVector& out) {
 }
 
 void allreduce(const detail::AllreduceOptionsImpl& opts) {
+  if (opts.elements == 0) {
```
SciPioneer, line 98: Shall we remove some sanity checks below, Such as `GLOO_ENFORCE(opts.elements > 0);` and `GLOO_ENFORCE(opts.elementSize > 0);`?

SciPioneer (approved):

gloo/allreduce.cc
agolynski: elementSize check should stay there i think. 

removed `GLOO_ENFORCE(opts.elements > 0)` since `opts.elements` is of size_t

SciPioneer: LGTM

PR #310, opened 2021-09-14
[ROCm] Changes not to rely on CUDA_VERSION or HIP_VERSION and use GLOO_USE_ROCM

pruthvistony:
- Changes to control hipify of CUDA_VERSION to HIP_VERSION
- use GLOO_USE_ROCM instead of __HIP_PLATFORM_HCC__
- Adding  __HIP_PLATFORM_AMD__ since  __HIP_PLATFORM_HCC__ is being deprecated.


pruthvistony:
@jeffdaily , @sunway513 and @jithunnair-amd
Please review these changes.

PR #313, opened 2021-12-08
Fix TSAN issue in gloo

pritamdamania87:
Summary:
The previous code was using `return map[name]`, which on the surface
seems like a read operation. However, std::map's `[]` operator would write a
default constructed value to the map if `name` was not found. This resulted in
concurrent unexpected write requests.

To resolve this issue, changed the logic to be read-only.

Differential Revision: D32938695


osalpekar (approved):
LGTM

jiayisuse (approved):
Thanks for the fix!

PR #320, opened 2022-03-05
Fix TSAN issue in gloo shutdown.

pritamdamania87:
Summary:
The problem arose from the fact that Loop::run() would have raw
pointers to Pair objects (Handler). As a result, as part of the destruction of
the process group the Pair would get destroyed whereas the loop would still try
to access these destroyed objects via a raw pointer.

To fix the issue, we now ensure the pairs are appropriately closed and the
device loop is stopped before any objects are destroyed.

Differential Revision: D34665049


osalpekar (approved):
Thanks for the fix!

PR #325, opened 2022-04-14
build rccl module only when USE_RCCL is ON

minsii:
Summary:
Current code always builds nccl module as long as USE_ROCM is ON.
However, we should build it only when USE_RCCL is ON. This patch fixes
it by adding the hipified nccl files in nccl/CMakeLists.txt which will be included
only when USE_RCCL is ON.

Differential Revision: D35636872


jiayisuse:
Thanks for fixing! Could we re-run the failed test at https://app.circleci.com/pipelines/github/facebookincubator/gloo/520/workflows/e701bbf6-cd61-4cdf-a38c-8903f83ef158/jobs/4942

PR #329, opened 2022-05-09
bad default constructor

zhouronghua:
compiled failed with clang:
    default constructor of 'Exception' is implicitly deleted because
    base class 'std::runtime_error' has no default constructor

    note: default constructor of 'InvalidOperationException' is implicitly
    deleted
    because base class '::gloo::Exception' has a deleted default constructor

  default constructor of 'IoException' is implicitly deleted because
  base class '::gloo::Exception' has a deleted default constructor


zhouronghua:
"https://[code.facebook.com](https://code.facebook.com/cla)/cla" can not be opened

PR #337, opened 2022-08-29
Fix build failure with clang-14

PhilArmstrong:
Need to explicitly include array header.

### DIFF
--- a/gloo/transport/tcp/device.cc
+++ b/gloo/transport/tcp/device.cc
@@ -12,6 +12,7 @@
 #include <netdb.h>
 #include <netinet/in.h>
 #include <string.h>
+#include <array>
 
 #include "gloo/common/linux.h"
 #include "gloo/common/logging.h"


PhilArmstrong:
Note that you also need to remove the default constructors from gloo/common/error.h as this constructor doesn’t exist in the std::runtime_error class & clang-14 will bomb out here.

As these are trivial changes, please feel free to make them yourselves, I don’t think I’ll be signing the CLA for these one-liners!

xgupta:
> Note that you also need to remove the default constructors from gloo/common/error.h as this constructor doesn’t exist in the std::runtime_error class & clang-14 will bomb out here.
> 
> As these are trivial changes, please feel free to make them yourselves, I don’t think I’ll be signing the CLA for these one-liners!

Then, you may close PR as a commit is made in the repo - https://github.com/facebookincubator/gloo/commit/4a5e339b764261d20fc409071dc7a8b8989aa195.

PR #346, opened 2022-11-01
Reorder structure for compatibility with linux-6.0

cdluminate:
Fixes: https://github.com/facebookincubator/gloo/issues/345
See also: https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=1022319

### DIFF
--- a/gloo/common/linux.cc
+++ b/gloo/common/linux.cc
@@ -193,8 +193,8 @@ static int getInterfaceSpeedGLinkSettings(int sock, struct ifreq* ifr) {
 #if LINUX_VERSION_CODE >= KERNEL_VERSION(4,6,0)
   constexpr auto link_mode_data_nwords = 3 * 127;
   struct {
-    struct ethtool_link_settings req;
     __u32 link_mode_data[link_mode_data_nwords];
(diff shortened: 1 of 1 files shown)


kumpera (approved):
LGTM

maurimo:
How can this be correct? the order (req first, the buffer after in the struct) is indeed to make space for a zero-sized array at the end of req. If you put req at the end then there is not enough memory and buffer might overflow?
See https://github.com/facebookincubator/gloo/pull/348 for an alternative fix to this issue.

Flamefire:
> How can this be correct? the order (req first, the buffer after in the struct) is indeed to make space for a zero-sized array at the end of req. If you put req at the end then there is not enough memory and buffer might overflow? See #348 for an alternative fix to this issue.

This change is indeed wrong. However it doesn't overflow the buffer (the size is still the same & `link_mode_data` is never touched) it will "just" not work at all: `ecmd.req.cmd = ETHTOOL_GLINKSETTINGS;` actually sets an entry in `link_mode_masks` (of `ethtool_link_settings`) and leaves the `cmd` zeroed which I guess in practice means the ioctl will always return an error.

PR #347, opened 2022-11-25
Fixed the bug in gatherv

WanXiaopei:
1. gatherv.h is not installed by the command of `make install`
2. in gloo/gatherv.cc, `GLOO_ENFORCE_GE` is a wrong check, for example: there are two tasks, A task send 3 and 4 elements to task A and task B, respectively, while task B send 5 and 6 elements to task A and task B, respectively. In this way, the `opts.elementsPerRank` of task A will be 3 and 5, apparently, 4 (the number of elements send to task B) is not greater equal than 5 (the number of elements receive from task B). 

1. gatherv.h在执行make install 时并没有拷贝到对应的目录，导致不能使用
2. 在gloo/gatherv.cc的107行和108行中，GLOO_ENFORCE_GE在这里的校验是错误的。比如，如果有两个任务，任务A分别向任务A和任务B发送3和4个数据，任务B分别向任务A和任务B发送5个和6个数据。那么对于任务A，则其opts.elementsPerRank将会是3和5，显然4（发送给任务B的数据量）并不是大于等于5（从任务B中收到的数据量）的。

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,2 +1,4 @@
 /build*
 *.pyc
+.vscode/
+.idea/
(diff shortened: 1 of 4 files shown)


WanXiaopei:
I do not know why `cuda11.7-all-transports` failed, is there any suggest ?

PR #348, opened 2023-01-13
Fix "flexible array member X not at end of struct" compiling with GCC-11

maurimo:
Fix a problem compiling with GCC-11 (present when I compiled latest Torch which includes gloo). The problem was also reported on Debian's ML, see for full context:
https://www.mail-archive.com/debian-bugs-dist@lists.debian.org/msg1875386.html

This problem should not be occurring in the first place and it's probably a bug of GCC, but luckily the workaround is quite simple too.

### DIFF
--- a/gloo/common/linux.cc
+++ b/gloo/common/linux.cc
@@ -192,29 +192,28 @@ const std::string& infinibandToBusID(const std::string& name) {
 static int getInterfaceSpeedGLinkSettings(int sock, struct ifreq* ifr) {
 #if LINUX_VERSION_CODE >= KERNEL_VERSION(4,6,0)
   constexpr auto link_mode_data_nwords = 3 * 127;
-  struct {
-    struct ethtool_link_settings req;
-    __u32 link_mode_data[link_mode_data_nwords];
-  } ecmd;
+  constexpr auto bufsize = sizeof(struct ethtool_link_settings) + sizeof(__u32)*link_mode_data_nwords;
+  char buf[bufsize];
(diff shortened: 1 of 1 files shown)


gloo/common/linux.cc
```diff
@@ -192,29 +192,28 @@ const std::string& infinibandToBusID(const std::string& name) {
 static int getInterfaceSpeedGLinkSettings(int sock, struct ifreq* ifr) {
 #if LINUX_VERSION_CODE >= KERNEL_VERSION(4,6,0)
   constexpr auto link_mode_data_nwords = 3 * 127;
-  struct {
-    struct ethtool_link_settings req;
-    __u32 link_mode_data[link_mode_data_nwords];
-  } ecmd;
+  constexpr auto bufsize = sizeof(struct ethtool_link_settings) + sizeof(__u32)*link_mode_data_nwords;
+  char buf[bufsize];
+  struct ethtool_link_settings& ecmd = *(struct ethtool_link_settings*)buf;
```
Flamefire, line 197: IMO this is technically a strict aliasing violation and hence UB. At least `buf` needs to be aligned as `struct ethtool_link_settings` and possibly a placement new needs to be used.

Idea: Using an union to enforce the size might work:
```
  constexpr auto bufsize = sizeof(struct ethtool_link_settings) + sizeof(__u32)*link_mode_data_nwords;
  union {
    struct ethtool_link_settings req;
    char buf[bufsize];
  } ecmd;
```

Flamefire (commented):
This is indeed correct while #346 is wrong. Needs rebase/Fix after that got already merged though.

CC @kumpera

PR #356, opened 2023-04-20
fix allgather for empty buffers (facebookincubator#179)

clumsy:
I used PyTorch-Ignite metrics that were all-gathered on both NCCL and Gloo backends.
In one of the cases the tensor (buffer) was empty and while NCCL tolerated this (hopefully as a no-op), Gloo gave me SIGFPE(8). I think this is because of the modulo by `dataSize` present today in Gloo all-gather code paths.
I believe this can be a no-op for Gloo, especially since `contextSize=1` (rank) is already a no-op.

### DIFF
--- a/gloo/allgather_ring.h
+++ b/gloo/allgather_ring.h
@@ -55,6 +55,10 @@ class AllgatherRing : public Algorithm {
   virtual ~AllgatherRing() {}
 
   void run() {
+    // Short circuit if there is only a single process or the output is empty.
+    if (this->contextSize_ == 1 || count_ == 0) {
+      return;
+    }
     const int rank = this->contextRank_;
     const int numRounds = this->contextSize_ - 1;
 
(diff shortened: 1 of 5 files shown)


gloo/allgather_ring.h
```diff
@@ -55,6 +55,10 @@ class AllgatherRing : public Algorithm {
   virtual ~AllgatherRing() {}
 
   void run() {
+    // Short circuit if there is only a single process or the output is empty.
+    if (this->contextSize_ == 1 || count_ == 0) {
```
malfet, line 59: Why `contextSize_` needs `this->`, but not `count_`?

malfet (approved):

gloo/allgather_ring.h
clumsy: That also struck me as unusual but I followed existing code that does it the same way, e.g. at line 71 in the original version

PR #357, opened 2023-04-26
[Store] Introduce optional append method.

kumpera:
Append is an optional operation for stores.
The semantics is to create or append the value to a given key.

Test plan:

It builds.
This change was tested from PyTorch.


fduwjj (commented):
Are existing test plans enough?

kumpera:
> Are existing test plans enough?

Yes, because we're not modifying exiting functionality.

The only implementations of the new store API will be in PT, so there isn't much to test here.

PR #359, opened 2023-05-10
fix test race condition

xunnanxu:
Summary: this would segfault if tests throw

Reviewed By: bmaurer

Differential Revision: D45723214


xunnanxu:
Already landed as 91c37366bab5248372a95351a82be58f4b679f63

PR #354, opened 2023-03-28
Add #include <exception> for std::exception_ptr

MaskRay:
Without the patch the file fails to compile with latest libc++ (https://reviews.llvm.org/D146097).

PR #375, opened 2023-10-26
Add `#include <exception>` to transport/tcp/unbound_buffer.h

hawkinsp:
This file uses `std::exception_ptr`, but fails to include the header that defines it, causing build failures in some enviroments.


hawkinsp:
Closing in favor of #354 , I missed that patch had already been sent. Could it please be merged?

PR #381, opened 2024-01-23
Add missing isConnected override to uv transport

dsciebu:
Project does not build with -DUSE_LIBUV=ON , due to the introduction of virtual isConnected() in base Pair class. This patch fixes it.


dsciebu:
> Thank you for the fix

np

Issue #345, opened 2022-10-25
fails to build against linux 6.0.3


cdluminate:
```
In file included from /<<PKGBUILDDIR>>/gloo/common/linux.cc:15:
/usr/include/linux/ethtool.h: In function ‘int gloo::getInterfaceSpeedGLinkSettings(int, ifreq*)’:
/usr/include/linux/ethtool.h:2088:17: error: flexible array member ‘ethtool_link_settings::link_mode_masks’ not at end of ‘struct gloo::getInterfaceSpeedGLinkSettings(int, ifreq*)::<unnamed>’
 2088 |         __u32   link_mode_masks[];
      |                 ^~~~~~~~~~~~~~~
/<<PKGBUILDDIR>>/gloo/common/linux.cc:197:11: note: next member ‘__u32 gloo::getInterfaceSpeedGLinkSettings(int, ifreq*)::<unnamed struct>::link_mode_data [381]’ declared here
  197 |     __u32 link_mode_data[link_mode_data_nwords];
      |           ^~~~~~~~~~~~~~
/<<PKGBUILDDIR>>/gloo/common/linux.cc:195:10: note: in the definition of ‘struct gloo::getInterfaceSpeedGLinkSettings(int, ifreq*)::<unnamed>’
  195 |   struct {
      |          ^
make[3]: *** [gloo/CMakeFiles/gloo.dir/build.make:303: gloo/CMakeFiles/gloo.dir/common/linux.cc.o] Error 1
```

Things are working for 5.X~5.19, but for linux 6.0.X it suddenly got broken.

PR #382, opened 2024-03-21
Fix wrong struct layout in getInterfaceSpeedGLinkSettings

Flamefire:
The `SIOCETHTOOL` IOCTL expects a pointer to an instance of `ethtool_link_settings`.
E.g. it will read the `cmd` member to determine what to do. However #346 reordered the memory layout of the pointer such that actually and array of `__32` values (zeroed out) is passed. Hence the IOCTL will either fail because an invalid command (`cmd=0`) is passed or the values read later by e.g.
`ecmd.req.link_mode_masks_nwords` are something completely different.

So `ethtool_link_settings` has to come before the (stack) memory used in the flexible array at the end of this struct.
To avoid GCC warnigns/errors (see #345) an union is used that provides the current struct (i.e. with wrong order) and an access the actually used `struct ethtool_link_settings` at the top.

Alternative approach to https://github.com/facebookincubator/gloo/pull/348 likely avoiding any undefined behavior (union based type punning is usually supported by compilers, so this should work to get the required memory)

### DIFF
--- a/gloo/common/linux.cc
+++ b/gloo/common/linux.cc
@@ -192,9 +192,12 @@ const std::string& infinibandToBusID(const std::string& name) {
 static int getInterfaceSpeedGLinkSettings(int sock, struct ifreq* ifr) {
 #if LINUX_VERSION_CODE >= KERNEL_VERSION(4,6,0)
   constexpr auto link_mode_data_nwords = 3 * 127;
-  struct {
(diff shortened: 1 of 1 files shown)


Flamefire:
@kumpera Would you be able to review and merge this? CI seems to be broken/unusable right now.

@cdluminate If you want/can take a look if this works for you too?

PR #402, opened 2024-12-12
Add CI for building on linux

c-p-i-o:
1. Update libuv version
2. Install openssl 1.1
Gloo is currently pinned to openssl 1.1.
See: gloo/test/CMakeLists.txt
It has:
find_package(OpenSSL 1.1 REQUIRED EXACT)
3. Build with c++17 to resolve libuv errors.
4. Always install libssl-dev
5. Add LD_LIBRARY_PATH for runners
6.  SSL max limit is 512M - 1

Also, make a small change to one of the tests.
SSL max limit is 512M - 1 byte
broadcast test fails at 512M, passes with 1 less byte.
TODO: comment where this limit is coming from.

### DIFF
--- a/.github/workflows/build_linux.yaml
+++ b/.github/workflows/build_linux.yaml
@@ -1,0 +1,68 @@
+name: CI-Linux
+on:
+  push:
+    branches:
+      - main
+  pull_request:
+    branches:
+      - main
+
+jobs:
+  build-linux:
+    name: build on ubuntu
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        cmake_args: ["", "-DUSE_IBVERBS=ON -DUSE_LIBUV=ON -DUSE_TCP_OPENSSL_LINK=ON", "-DUSE_TCP_OPENSSL_LOAD=ON"]
+
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v4
+      - name: Setup environment
+        run: |
+          sudo apt-get install -y libssl-dev
+      - name: Install apt packages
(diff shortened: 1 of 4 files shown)


.github/workflows/build_linux.yaml
```diff
@@ -0,0 +1,35 @@
...
+    strategy:
+      matrix:
+        openssl: [true, false]
+
+    steps:
+      - name: Checkout code
+        uses: action/checkout@v4
+      - name: Setup environment
+        run: |
+          if [ "$OPENSSL" = "true" ]; then
```
malfet, line 23: PR is missing 
```
   env:
    OPENSSL: ${matrix.openssl}
```
or something of this nature

.github/workflows/build_linux.yaml
```diff
@@ -0,0 +1,35 @@
...
+            sudo apt-get install -y libssl-dev
+          fi
+      - name: Install libuv
+        run: |
+          apt-get install -y wget pkg-config
+          wget -q -O libuv-v1.26.0.tar.gz https://dist.libuv.org/dist/v1.26.0/libuv-v1.26.0.tar.gz
```
malfet, line 29: 1.26 is from 2019, and I think we build everything in PT using 1.42 or something like that, so please update

c-p-i-o: Thanks! Yeah - just copying over from circleci to revive the CI at the moment and getting the build going. Will update to latest!

fduwjj (approved):

PR #406, opened 2025-02-07
Formatting changes + fix windows OSS build

c-p-i-o:
Summary: Add .clang-format 
Copied from https://github.com/pytorch/pytorch/blob/main/.clang-format

Differential Revision: D69275930

### DIFF
--- a/.github/workflows/build-windows.yml
+++ b/.github/workflows/build-windows.yml
@@ -49,22 +49,22 @@ jobs:
           cd libuv-v1.38.0
           mkdir -p build
           cd build
-          mkdir -p ${{ env.gtest_lib_path }}
+          mkdir -p ${{ env.libuv_path }}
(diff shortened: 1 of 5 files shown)


wconstab (approved):
stamp to unblock.
nit: why 2 copies of clang-format?
also, i didn't really review clang-format, i assume you copied it from somewhere and the configs are sane?  (maybe say where its copied from in the comment, but nbd)

c-p-i-o:
> stamp to unblock. nit: why 2 copies of clang-format? also, i didn't really review clang-format, i assume you copied it from somewhere and the configs are sane? (maybe say where its copied from in the comment, but nbd)

copied from pytorch.

somehow the github workflows weren't able to ref the .clang-format in root directory so made a copy.
will tackle this separately.

fduwjj (approved):

PR #411, opened 2025-02-17
[Feat]add tcp store for rendezvous

konnase:
1. add tcp store for rendezvous

usage:
```c++
auto tcp_store = std::make_shared<gloo::rendezvous::TCPStore>("127.0.0.1", 12345, 2, is_master);
```


d4l3k:
@konnase what's your usecase for this? How are you currently using Gloo?

We typically use PyTorch's TCPStore for the Gloo integration in PT.

konnase:
We need to use ibverbs in gloo, but pytorch do not support ibverbs with gloo backend, so we want to use gloo directly. We found gloo does not have tcp store, so we want to implement a tcp store in gloo.

d4l3k:
@konnase we now have initial support for the ibverbs backend in PyTorch -- still has some bugs we're working out though

PR #416, opened 2025-03-08
Fix missing header for getenv

cyyever:
getenv can't be found on MacOS Apple Silicon.

### DIFF
--- a/gloo/common/utils.cc
+++ b/gloo/common/utils.cc
@@ -11,7 +11,7 @@
 #ifdef _WIN32
 #include <winsock2.h>
 #else
-#include <stdlib.h>
+#include <cstdlib>
 #include <unistd.h>
 #endif
 
--- a/gloo/common/utils.cc
+++ b/gloo/common/utils.cc
@@ -11,8 +11,8 @@
 #ifdef _WIN32
 #include <winsock2.h>
 #else
-#include <cstdlib>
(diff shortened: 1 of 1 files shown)


gloo/common/utils.cc
```diff
@@ -11,6 +11,7 @@
 #ifdef _WIN32
 #include <winsock2.h>
 #else
+#include <stdlib.h>
```
pganssle-google, line 14: On Linux with `clang`, I find that this only works if I use `<cstdlib>` here.

```suggestion
#include <cstdlib>
```

If that also works in your environment, I would recommend changing it?

malfet (approved):

gloo/common/utils.cc
malfet: This should be the case, let's use it

d4l3k (approved):

fduwjj (approved):

atalman (approved):
lgtm

gloo/common/utils.cc
```diff
@@ -11,6 +11,7 @@
 #ifdef _WIN32
 #include <winsock2.h>
 #else
+#include <cstdlib>
 #include <unistd.h>
```
malfet, line 15:
```suggestion
#include <unistd.h>
#include <cstdlib>
```

PR #422, opened 2025-03-28
gloo/transport/tcp/loop: better memory management and shutdown

d4l3k:
Summary:
This overhauls how we schedule async operations on the tcp Loop. 

These changes greatly simplifies the memory management and simplifies the shutdown logic since memory ownership is better tied to epoll usage and relies less on shared_ptr copies everywhere to work around it.

This makes a few major changes:

* adds a new `registerDescriptor` method to Loop that provides a shared_ptr. This method will track the lifetime of the provided Handler and will release it when the epoll handle is updated/deleted.
* Make `Handler::handleEvents` signature provide a `Loop&` reference so all objects don't need to hold a shared_ptr to the loop.
* Update `helpers.h` to not manage it's own lifetime (instead using registerDescriptor shared_ptr support) and also not hold a shared_ptr to `Loop`.
* Introduces a new `shutdown()` method that allows the loop to gracefully shutdown all operations prior to destruction. This simplifies teardown on error since there a few places where handlers (Device) are managed with raw pointers. This isn't used outside of Loop currently but will be in a follow up PR.

Differential Revision: D72080276

### DIFF
--- a/gloo/test/tcp_test.cc
+++ b/gloo/test/tcp_test.cc
@@ -17,7 +17,7 @@ TEST(TcpTest, ConnectTimeout) {
   // Use bad address
   auto remote = Address("::1", 10);
   auto timeout = std::chrono::milliseconds(100);
-  auto fn = [&](std::shared_ptr<Socket>, const Error& e) {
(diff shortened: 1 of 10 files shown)


fduwjj (approved):
LGTM

PR #424, opened 2025-03-28
Update CMake version

malfet:
As cmake-4.0.0 is no longer compatible with cmake-2, but those files are solid cmake-3.x project


d4l3k (approved):
LGTM

PR #428, opened 2025-04-11
Include cstdlib header in common/utils.cc

pganssle-google:
This is required to use `getenv`.


pganssle-google:
Oops, this is essentially a duplicate of #416

PR #429, opened 2025-04-16
gloo/tcp/attr: hold on to bound socket to avoid port conflict issues

d4l3k:
Summary:
This is a minor refactor to attr in the TCP backend for Gloo so that we don't close the bound fd between allocating it in `lookupAddrForHostname` and binding it in `Listener`. 

Listener sets `SO_REUSEADDR` so it's fine for us to bind twice to the same address but we will only listen on it in `Listener`.

Differential Revision: D73148977


fduwjj (approved):
thanks for the fix.

PR #434, opened 2025-04-28
Doing the proper error check after calling getaddrinfo

sampathvic:
Summary:
Function `getaddrinfo()` is expected to return 0 on success and other error codes in case of failures as per the man page: https://fburl.com/tg3q11nw. However, current logic in `lookupAddrForHostname()` function is just checking for return value not equal to -1: https://fburl.com/code/wk7ztua9 for error check.This fix ensures that the return value of `getaddrinfo()` zero on suceessfull cases.

 {F1977470144}

Differential Revision: D73746837


d4l3k (approved):
LGTM

PR #435, opened 2025-05-02
gloo: use unbound buffers for barrier, broadcast

d4l3k:
Summary:
This makes barrier and broadcast operations in Gloo use UnboundBuffer instead of Buffer. UnboundBuffer is the newer preferred buffer implementation and also makes the algorithms much easier to express/understand and efficient since we don't need to allocate more memory.

Tracking issue for Buffer removal: https://github.com/pytorch/gloo/issues/432

Differential Revision: D74096980

### DIFF
--- a/gloo/broadcast_one_to_all.h
+++ b/gloo/broadcast_one_to_all.h
@@ -36,32 +36,6 @@ class BroadcastOneToAll : public Algorithm {
     GLOO_ENFORCE_LT(rootRank_, contextSize_);
     GLOO_ENFORCE_GE(rootPointerRank_, 0);
     GLOO_ENFORCE_LT(rootPointerRank_, ptrs_.size());
-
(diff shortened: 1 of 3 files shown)


fduwjj (approved):
LGTM, this is way easier to understand, I just read the unbound buffer and it way easier to understand than the original. And it is super laborious to do a lot of set up code when using buffer previously.

gloo/broadcast_one_to_all.h
```diff
@@ -70,14 +44,21 @@ class BroadcastOneToAll : public Algorithm {
       return;
     }
 
+    auto clearToSendBuffer = context_->createUnboundBuffer(nullptr, 0);
```
fduwjj, line 47: N00b question why this is even needed? Is it because we need to wait for the receiver to be ready?

PR #437, opened 2025-05-07
c10d/gloo: add ibverbs backend

d4l3k:
Summary:
This provides a new "UnboundBuffer" implementation for Gloo ibverbs backend so it can be used with PyTorch.

This currently is passing basic tests such as `reduce_test` and `send_recv_test` but there are a number of failures. Putting this up for review so the follow up fixes are less of a mega PR and also so we can start doing some initial testing with this E2E with PyTorch.

Known issues:

* support recv from any is not supported
* AllreduceBcubeBase2 is failing

Differential Revision: D73291471


gloo/transport/ibverbs/buffer.cc
```diff
@@ -59,8 +59,28 @@ Buffer::Buffer(Pair* pair, int slot, void* ptr, size_t size)
...
+    std::unique_lock<std::mutex> lock(m_);
+    sendPending = sendPending_;
+  }
+
+  if (sendPending > 0) {
+    GLOO_WARN(
+        "Destructing buffer with pending sends, sendPending_=", sendPending_);
+    try {
+      for (int i = 0; i < sendPending; i++) {
+        waitSend();
```
fduwjj, line 73: Is this blocking so the destructor can hang?

gloo/transport/ibverbs/device.cc
```diff
@@ -75,16 +75,31 @@ std::shared_ptr<::gloo::transport::Device> CreateDevice(
...
+          devices[i]->name,
+          " dev=",
+          devices[i]->dev_name);
       names.push_back(devices[i]->name);
     }
     std::sort(names.begin(), names.end());
     attr.name = names[0];
   }
 
+  attr.index = 3;
```
fduwjj, line 89: N00b question, why this magic number? Is it possible that we can add a comment for it?

gloo/transport/ibverbs/unbound_buffer.cc
```diff
@@ -0,0 +1,257 @@
...
+UnboundBuffer::~UnboundBuffer() {
+  std::lock_guard<std::mutex> guard(m_);
+
+  ibv_dereg_mr(mr_);
+}
+
+void UnboundBuffer::abortWaitRecv() {
+  std::lock_guard<std::mutex> guard(m_);
+  abortWaitRecv_ = true;
+  recvCv_.notify_one();
```
fduwjj, line 85: IIUC, this does guarantee which thread to wakes up. Would this be a problem? Oh maybe it is per pair, so it won't be a problem?

gloo/transport/ibverbs/pair.cc
```diff
@@ -21,9 +22,11 @@ namespace transport {
 namespace ibverbs {
 
 Pair::Pair(
+    int rank,
```
fduwjj, line 25: To make sure my understand correctly, this is always global rank right? For example, for Sub-PG scenario, we sometimes will have local ranks as well.

gloo/transport/ibverbs/pair.cc
```diff
@@ -238,12 +248,13 @@ void Pair::sendMemoryRegion(struct ibv_mr* src, int slot) {
...
   // They are posted in FIFO order, but may complete in arbitrary order.
   // Therefore we store them in a map keyed on the buffer slot.
-  GLOO_ENFORCE_EQ(mappedSendRegions_.count(slot), 0);
-  mappedSendRegions_[slot] = std::move(mr);
+  mappedSendRegions_[slot].emplace_back(std::move(mr));
 }
 
-const struct ibv_mr* Pair::getMemoryRegion(int slot) {
-  std::unique_lock<std::mutex> lock(m_);
+void Pair::getMemoryRegion(
```
fduwjj, line 254: IMHO, If we now return nothing, do you think we still name it `getMemoryRegion`. I mean when we call getSomething, we expect something will be returned?

gloo/transport/ibverbs/pair.cc
```diff
@@ -301,41 +303,124 @@ void Pair::postReceive() {
 
 std::unique_ptr<::gloo::transport::Buffer>
 Pair::createSendBuffer(int slot, void* ptr, size_t size) {
+  // GLOO_THROW("createSendBuffer not supported for ibverbs transport");
```
fduwjj, line 306: nit: remove?

gloo/transport/ibverbs/pair.cc
```diff
@@ -301,41 +303,124 @@ void Pair::postReceive() {
...
   auto buffer = new Buffer(this, slot, ptr, size);
-  sendCompletionHandlers_[slot] = buffer;
+  auto& q = sendCompletionHandlers_[slot];
+  q.clear();
+  q.emplace_back(buffer);
+
+  // getMemoryRegion(lock, slot, [](struct ibv_mr peer) {
+  //   GLOO_DEBUG("createSendBuffer peer_addr=", peer.addr, " rkey=",
+  //   peer.rkey);
+  // });
```
fduwjj, line 318: ditto

gloo/transport/ibverbs/pair.cc
```diff
@@ -301,41 +303,124 @@ void Pair::postReceive() {
...
+
   return std::unique_ptr<::gloo::transport::Buffer>(buffer);
 }
 
 std::unique_ptr<::gloo::transport::Buffer>
 Pair::createRecvBuffer(int slot, void* ptr, size_t size) {
+  // GLOO_THROW("createRecvBuffer not supported for ibverbs transport");
```
fduwjj, line 325: ditto

gloo/transport/ibverbs/pair.cc
```diff
@@ -301,41 +303,124 @@ void Pair::postReceive() {
...
-      "Unbound buffers not supported yet for ibverbs transport");
+    uint64_t slot,
+    size_t offset,
+    size_t nbytes) {
+  std::unique_lock<std::mutex> lock(m_);
+  GLOO_DEBUG("send tag=", slot, " offset=", offset, " nbytes=", nbytes);
+
+  GLOO_ENFORCE(!sync_, "Cannot send in sync mode");
+  // GLOO_ENFORCE_EQ(offset, 0);
+  //  GLOO_ENFORCE_EQ(nbytes, tbuf->size);
```
fduwjj, line 348: ditto

gloo/transport/ibverbs/pair.cc
```diff
@@ -301,41 +303,124 @@ void Pair::postReceive() {
...
+    uint64_t tag,
+    size_t offset,
+    size_t nbytes) {
+  std::unique_lock<std::mutex> lock(m_);
+
+  GLOO_DEBUG("recv tag=", tag, " offset=", offset, " nbytes=", nbytes);
+
+  GLOO_ENFORCE(!sync_, "Cannot recv in sync mode");
+  // GLOO_ENFORCE_EQ(offset, 0);
+  // GLOO_ENFORCE_EQ(nbytes, tbuf->size);
```
fduwjj, line 414: ditto

gloo/transport/ibverbs/pair.cc
fduwjj: After reading L360, maybe we can call it recv or fetchMemoryRegion? No strong opinion..

gloo/transport/ibverbs/pair.cc
```diff
@@ -478,43 +607,67 @@ void Pair::handleCompletion(struct ibv_wc* wc) {
...
-    GLOO_ENFORCE_GT(mappedSendRegions_.size(), 0);
-    GLOO_ENFORCE_EQ(mappedSendRegions_.count(slot), 1);
-    mappedSendRegions_.erase(slot);
+    mappedSendRegions_[slot].pop_front();
   } else {
     GLOO_ENFORCE(false, "Unexpected completion with opcode: ", wc->opcode);
   }
 }
 
 void Pair::send(Buffer* buffer, size_t offset, size_t length, size_t roffset) {
```
fduwjj, line 616: another n00b question, why there are two send? Is it possible that we can have comments to explain under what condition this send will be used? For example the comment of the first one says "// Send from the specified buffer to remote side of pair." I guess this one is similar. Even the code looks similar.

fduwjj (approved):
Great work. Aside from my n00b questions, it looks good to me overall.

fduwjj (approved):
Great work. Aside from my n00b questions, it looks good to me overall.

gloo/transport/ibverbs/pair.cc
d4l3k: It's for the two different types of buffers. Buffer is bound to a specific pair and reuses the memory region where as UnboundBuffer can be used across multiple pairs and will bind the memory region just for that transfer.

I tried to generalize this logic but had to split it given the differences in apis between the two

gloo/transport/ibverbs/pair.cc
d4l3k: done

gloo/transport/ibverbs/pair.cc
d4l3k: This is the rank of the Gloo process group/context. On the gloo side there's only a single rank concept per context, no understanding of sub PG ranks etc.

gloo/transport/ibverbs/device.cc
d4l3k: Removed -- this is infra specific, should be passing it in via the attr and not hardcoding

gloo/transport/ibverbs/unbound_buffer.cc
d4l3k: abortWaitRecv only aborts a single wait request so we only want to wake one thread. I don't think we ever call multiple waitRecv on the same buffer. Usually each algorithm is only waiting for one thing at a time since it's a blocking call

gloo/transport/ibverbs/buffer.cc
d4l3k: This actually isn't needed anymore -- there was an issue that's since been fixed. Changed this to just be a warning now

gloo/transport/ibverbs/pair.cc
fduwjj: Got it, thanks for your explanation.

PR #439, opened 2025-05-13
Add .clang-format config, provide to super-linter, and apply formatting to codebase

jackbondpreston-arm:
55b6bf777b68a9c36b7034ca615411cb8d0cab15 applied clang-format to the codebase, but did not add the configuration file used. This change adds the configuration file, `.clang-format`, and applies clang-format to the codebase. The configuration file is imported from [PyTorch](https://github.com/pytorch/pytorch), with attribution in comments at the beginning of the file.

The config file used seems to line up well with the one used for 55b6bf777b68a9c36b7034ca615411cb8d0cab15, as the diff generated is relatively small. Having the config available will enable contributions going forward to use consistent formatting and avoid having to apply it across the whole codebase again.

To further this aim, I have linked the `.clang-format` into `.github/config/lint`, to enable the `lint` job in CI to utilise it for formatting checking. Additionally, I have updated the version of `super-linter` used in this job to a newer one, as this provides a much newer version of `clang-format`.

I have verified the super-linter job locally, using [act](https://github.com/nektos/act), and have confirmed it seems to be working fine. For this reason, I suggest that we can re-enable the linting job in CI, to ensure contributions now follow the `.clang-format`.

The benefit to this project is that we can have a consistent and clear formatting style, saving time for developers and reviewers by avoiding having to think/argue about formatting - it can simply be applied with a standard tool.

### DIFF
--- a/.clang-format
+++ b/.clang-format
@@ -1,0 +1,168 @@
+# From https://github.com/pytorch/pytorch/blob/379bbef23c144aba4ae6e01d9e1247c4a44b3e74/.clang-format
+# License:
+# From PyTorch:
+# 
+# Copyright (c) 2016-     Facebook, Inc            (Adam Paszke)
+# Copyright (c) 2014-     Facebook, Inc            (Soumith Chintala)
+# Copyright (c) 2011-2014 Idiap Research Institute (Ronan Collobert)
+# Copyright (c) 2012-2014 Deepmind Technologies    (Koray Kavukcuoglu)
+# Copyright (c) 2011-2012 NEC Laboratories America (Koray Kavukcuoglu)
+# Copyright (c) 2011-2013 NYU                      (Clement Farabet)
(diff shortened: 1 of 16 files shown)


d4l3k (approved):
LGTM

Would be nice to setup a lintrunner CI job as well but this is a good start

jackbondpreston-arm:
> Would be nice to setup a lintrunner CI job as well but this is a good start

What does this mean, compared to the existing super-linter job?

d4l3k:
@jackbondpreston-arm Doesn't seem like the super-linter CI job is running? Need to follow up on that

jackbondpreston-arm:
> @jackbondpreston-arm Doesn't seem like the super-linter CI job is running? Need to follow up on that

Yes, I imagine it was disabled due to it previously not having any `.clang-format` configuration for clang-format linting. This would have caused tonnes of failures, as Gloo's formatting doesn't line up with clang-format's default config - which is likely why it got disabled. I addressed this issue in this patch, and confirmed that (at least locally, using `act`) the super-linter job is working as expected again.

So, I think we should be good to re-enable the `super-linter` job.

PR #440, opened 2025-05-13
remove OpenSSL requirement for testing

nlbrown2:
Enable building of gloo_test without requiring OpenSSL to be present. If OpenSSL is not present, SSL/TLS related tests are not linked. If OpenSSL is present, then TLS tests are compiled and linked.


d4l3k (approved):
LGTM

PR #446, opened 2025-05-19
[gloo] Enable using c10::Half for gloo cuda

fduwjj:
Stack from [ghstack](https://github.com/ezyang/ghstack) (oldest at bottom):
* __->__ #446

While debugging numerical issues for gloo when it comes to float16, we found that using original data type can lead to errors as is shown in https://github.com/pytorch/pytorch/issues/152300. So instead of digging into the rabbit hole of why the existing implementation is wrong, we think switching to c10::Half might be a better idea because the PyTorch version uses accelerated operations when available (CPU extensions or AVX) and supports both CPU and GPU.


d4l3k (approved):
LGTM

PR #458, opened 2025-07-31
Intra-node shared memory (SHM) optimizations for CPU primitives

gaopengff:
This PR is for RFC https://github.com/pytorch/gloo/issues/455. It has implemented shm allreduce.

### DIFF
--- a/gloo/allreduce_shm.cc
+++ b/gloo/allreduce_shm.cc
@@ -61,7 +61,7 @@ void shared_create(
     size_t nbytes) {
   int d = shm_open(name, O_CREAT | O_RDWR, S_IRUSR | S_IWUSR);
   if (d != -1) {
-    if (ftruncate(d, nbytes) != -1) {
+    if (nbytes = write(d, bytes, nbytes)) {
       shared_open(data, name, nbytes);
     }
     close(d);


gloo/allreduce_shm.cc
```diff
@@ -0,0 +1,443 @@
...
+  }
+
+  current_buffer = 1 - current_buffer;
+}
+
+} // namespace
+
+bool is_intra_node(const int size) {
+    // must launch with torchrun
+  auto local_size_string = std::getenv("LOCAL_WORLD_SIZE");
```
d4l3k, line 381: I don't think this check is safe -- for torchft for instance we often run with Gloo only cross host and if you're using an 8x8 configuration this would trigger shm logic for cross host comms

d4l3k: @gaopengff With hybrid parallelism we often create cross host groups that can be smaller than LOCAL_WORLD_SIZE.

I.e. with HSDP in a 2x8 configuration we could be using FSDP within one host and replicating across the hosts. Thus each worker has two instances of Gloo. Once that is size 8 for within the host and then the other with size 2 between each pair of ranks across hosts.

The opposite can happen as well. Say we're running with a tensor parallel dimension of 2 -- that means each 2 workers on one host will have a group of size 2 and this logic would incorrectly not use shm for that operation.

d4l3k (changes requested):
Still need to fix the triggering logic + add some unit tests for this behavior

gloo/allreduce_shm.cc
gaopengff: In this hybrid scenario, checking for LOCAL_WORLD_SIZE from environment variables is not correct indeed. It's difficult to get local_world_size in gloo context. So I changed the code to check **max_local_rank** whether is to equal to **world_size**. Supposing we are running on 8x8 configuration:
Group1: 1x8 intra-node max_local_rank(8) == world_size(8) -> shm allreduce
Group2: 8x1 inter-node max_local_rank(1) != world_size(8) -> ring allreduce
Do you think the check is right?

gloo/allreduce.cc
```diff
@@ -131,14 +132,23 @@ void allreduce(const detail::AllreduceOptionsImpl& opts) {
     return;
   }
 
-  switch (opts.algorithm) {
+  auto algorithm = opts.algorithm;
+  if (is_intra_node(opts)) {
```
d4l3k, line 136: Doing a broadcast prior to every single allreduce seems pretty expensive -- can we move this logic to the TCP init in `createAndConnectAllPairs`? We have enough information there I believe to compute the topology since we have all the hostnames.

d4l3k (changes requested):

gloo/allreduce.cc
d4l3k: The other option would be to inspect the pairs that are participating and check if they all share the same IP

gloo/allreduce.cc
```diff
@@ -131,14 +132,25 @@ void allreduce(const detail::AllreduceOptionsImpl& opts) {
     return;
   }
 
-  switch (opts.algorithm) {
+  auto algorithm = opts.algorithm;
+
+#ifndef _WIN32
+  if (context->isIntraNode()) {
```
d4l3k, line 138: Can we add a check to make sure inputs are on CPU before setting it to shm? I believe this would fail when using local ibv cuda backend

Can also check `device->hasGPUDirect()` and instead disable for all GPU direct supported backends

d4l3k (changes requested):

gloo/allreduce.cc
gaopengff: Sure. I've changed the check condition to "**(context->isIntraNode() && !context->getDevice()->hasGPUDirect())**".

d4l3k (approved):
LGTM

d4l3k:
@gaopengff looks like a legitimate failure: AllreduceNewRing/AllreduceNewTest

gaopengff:
Hi, @d4l3k  It seems that the recently CI failure "terminate called after throwing an instance of 'gloo::EnforceNotMet'" is not caused by this PR's change. Also, I have set up my owe CI environment to test this PR. According my own CI results at [link](https://github.com/gaopengff/gloo/actions/runs/19730536635), it should pass the Linux CI. Could you help retrigger it, thanks.

d4l3k:
This is causing issues due to breaking arm builds -- going to revert it for now and reland with the arm fixes.

PR #460, opened 2025-08-07
Fix Mac build failures

cyyever:


malfet (changes requested):
@cyyever can you provide some context (i.e. failure log for example)

PR #467, opened 2025-09-12
Add GLOO_LOG_LEVEL env

kwen2501:
Accepted values:
ERROR, WARN, INFO, DEBUG

Default value:
WARN


youkaichao (commented):
hope we can make it work soon, so that https://github.com/pytorch/gloo/pull/466 is useful

PR #484, opened 2026-01-20
Support async mode for shm allreduce

gaopengff:
This is to fix CI failure of pytorch in bump PR https://github.com/pytorch/pytorch/pull/172297.
In async mode **shmData** should be occupied exclusively. We added lock for **shmData** to make it thread safe and used unique tag to do synchronization among different ranks.


gaopengff:
@d4l3k Could you help review this?

d4l3k (approved):
LGTM

gloo/context.h
```diff
@@ -37,6 +37,8 @@ class Context {
 
   std::shared_ptr<AllreduceSharedMemoryData> shmData;
 
+  std::mutex shmDataMutex;
```
d4l3k, line 49: What happens if there's multiple gloo process groups? Does that cause issues at all?

d4l3k: also can we put this under shmData?

gaopengff: 1. For multiple process groups scenario, I used gloo context's address to generate unique ID for shm name. In that case different group will use different shm buffer to do allreduce op. I've verified with a test with multiple process groups with pytorch and it passed.
2. I think we could not put this under shmData. In the first run shmData is not initialized(nullptr), if there are multiple threads reaching this point, we need to ensure the initialization work is done only by one thread here.

gloo/allreduce.cc
```diff
@@ -131,14 +135,27 @@ void allreduce(const detail::AllreduceOptionsImpl& opts) {
     return;
   }
 
-  switch (opts.algorithm) {
+  auto algorithm = opts.algorithm;
+
+#if !defined(_WIN32) && !defined(__aarch64__) && !defined(__arm__)
+  if (context->isIntraNode() && !context->getDevice()->hasGPUDirect()) {
+    algorithm = detail::AllreduceOptionsImpl::SHM;
```
kapilsh, line 143: I dont see users to be able to use explicit algorithm - this will override anything user explicitly specifies. should we check `Algorithm::UNSPECIFIED` before we override?

gloo/context.h
kapilsh: > I think we could not put this under shmData. In the first run shmData is not initialized(nullptr), if there are multiple threads reaching this point, we need to ensure the initialization work is done only by one thread here.

We could potentially move this to allreduce_shm.cc and make both static? that way we keep global context clean of shm specifics. wdyt?

gloo/allreduce.cc
```diff
@@ -12,8 +12,12 @@
 #include <array>
 #include <cstring>
 
+#if !defined(_WIN32) && !defined(__aarch64__) && !defined(__arm__)
```
kapilsh: this seems to be copied in a bunch of places - can we make this a macro?

gloo/allreduce_shm.cc
```diff
@@ -0,0 +1,556 @@
...
+  memcpy(to_buffer + offset, buffers[0] + offset, num_elements * element_size);
+  for (int i = 1; i < world_size; i++) {
+    fn(to_buffer + offset,
+       to_buffer + offset,
+       buffers[i] + offset,
+       num_elements);
+  }
+}
+
+static void parallel_memcpy(void* to, void* from, size_t n_bytes)
```
kapilsh, line 161: for smaller buffers - this might not be optimal. may be we can just use memcpy and optimize with some benchmarks in a later PR?

kapilsh (changes requested):
Thanks looks great - I left a few comments. can we also add some tests for covering the new shm all_reduce and possibly resource cleanup tests for shm

gloo/allreduce.cc
gaopengff: I've modified it to make sure it will override **Algorithm::UNSPECIFIED** only when shm allreduce is applicable. Also I added unit test for shm allreduce in **gloo/test/allreduce_test.cc**

gloo/allreduce.cc
gaopengff: Sure, I've defined a macro of this in gloo/allreduce.h, which will be used in unit test too,
```C++
#if !defined(_WIN32) && !defined(__aarch64__) && !defined(__arm__)
#define GLOO_SHM_ALLREDUCE_APPLICABLE 1
#else
#define GLOO_SHM_ALLREDUCE_APPLICABLE 0
#endif
```

gloo/allreduce_shm.cc
gaopengff: Now I changed it to use memcpy when buffer size is less than 1MiB. Do you think it's OK?
```C++
  // use memcpy when buffer size is less than 1 MiB
  if (n_bytes < AllreduceSharedMemoryData::AllreduceWorkspace::
                    NAIVE_ALLREDUCE_THRESHOLD) {
    memcpy(to, from, n_bytes);
    return;
  }
```

gloo/context.h
gaopengff: It works only on multi-processes scenario if we make them static. However, in multi-threads scenario like gloo unit test, each thread will represent a rank and the shm_data is only initialized once, which is not as expected.
Although there is a keywork **thread_local** to make static variable unique among threads, it may cause performance issue. In real workload such as Pytorch the calling sequence is more like:
```bash
context init -> call allreduce -> call allreduce -> call allreduce
```
**thread_local** will make shm data initialized every time calling allreduce. The initialization is very expensive as it will allocate shm buffer.

kapilsh: > thread_local will make shm data initialized every time calling allreduce

is that correct? I think it will be initialized only the first time but I see your point. Okay I think we can at least wrap this in the macro you created.

gloo/allreduce_shm.cc
kapilsh: I think let's start with this. Are you able to run some benchmarks? fwiw it will platform and hardware specific. I am just curious whether we get any benefit at all from parallelization but with measurements hard to say.

kapilsh (approved):

gloo/context.h
gaopengff: OK, I've wrapped shm_data's declaration in the macro I created.

gaopengff:
Hi @kapilsh Is there other comments about this PR? The CI passed could we merge it?

kapilsh:
> Hi @kapilsh Is there other comments about this PR? The CI passed could we merge it?

Hey, there is failing test target - dont think related to your PR but we need to double check. we are looking at it and will merge as soon as that is resolved. thanks!

gaopengff:
> > Hi @kapilsh Is there other comments about this PR? The CI passed could we merge it?
> 
> Hey, there is failing test target - dont think related to your PR but we need to double check. we are looking at it and will merge as soon as that is resolved. thanks!

Hi, @kapilsh It's been a while

> > Hi @kapilsh Is there other comments about this PR? The CI passed could we merge it?
> 
> Hey, there is failing test target - dont think related to your PR but we need to double check. we are looking at it and will merge as soon as that is resolved. thanks!

Hi, just following up — it looks like the CI issue may be resolved now. Could you help retrigger CI for this PR? Thanks!

kapilsh:
hi @gaopengff , sorry for the delay on this. I chatted offline with @d4l3k - unfortunately this is breaking a bunch of internal tests - we are working on it. do you mind at least rebasing your changes so CI can trigger updated tests. i think the CI issue that you faced earlier should be resolved now.

gaopengff:
> hi @gaopengff , sorry for the delay on this. I chatted offline with @d4l3k - unfortunately this is breaking a bunch of internal tests - we are working on it. do you mind at least rebasing your changes so CI can trigger updated tests. i think the CI issue that you faced earlier should be resolved now.

OK, I just rebased my branch to the latest main.

gaopengff:
> > hi @gaopengff , sorry for the delay on this. I chatted offline with @d4l3k - unfortunately this is breaking a bunch of internal tests - we are working on it. do you mind at least rebasing your changes so CI can trigger updated tests. i think the CI issue that you faced earlier should be resolved now.
> 
> OK, I just rebased my branch to the latest main.

Just wondering, do you think this could still catch the PyTorch 2.12 release?

gaopengff:
Hi @kapilsh, about meta internal CI result of this PR, is there any update?

PR #487, opened 2026-02-05
ci: add arm64 runner for github actions

nlbrown2:
Extend existing Linux build workflow to also run on arm64 Ubuntu runners


d4l3k (approved):
LGTM

PR #488, opened 2026-02-06
ci: add CUDA and rocm builds

d4l3k:
As titled, this adds CUDA and rocm builds

This doesn't run any tests just confirms builds are successful.


amirafzali (approved):
lgtm

PR #491, opened 2026-02-16
CMake: Use HIPIFY_OUTPUT_DIR when installing ROCm headers

daissi:
When USE_ROCM is enabled, the install path for HIP headers was derived by replacing CMAKE_CURRENT_SOURCE_DIR in the header directory path.

However, HIPified headers are generated into HIPIFY_OUTPUT_DIR, not CMAKE_CURRENT_SOURCE_DIR. As a result, the computed install path is incorrect.

This change replaces CMAKE_CURRENT_SOURCE_DIR with HIPIFY_OUTPUT_DIR when rewriting header paths, ensuring ROCm headers are installed into the correct include/gloo/... location.

This fixes incorrect include directory layout when building with ROCm.

This [patch is applied for the Debian packaging](https://salsa.debian.org/deeplearning-team/gloo/-/blob/master/debian/patches/Fix_header_install.patch) otherwise headers are installed in the wrong location.


daissi:
cc @GZGavinZhao @d4l3k

GZGavinZhao (approved):
The change is reasonable to me. I'm curious though how the official ROCm PyTorch distribution was able to build/compile with the previous code. I assume it's because the official ROCm wheels have been built with gloo as a submodule so it doesn't have this problem...?

PR #517, opened 2026-06-26
expose remote peer rank in transport::Pair (#517)

amirafzali:
Summary:

Enhancing transport::Pair debugging with remote peer rank information (in addition to existing remote address)

Differential Revision: D109783094


d4l3k (approved):
LGTM

PR #518, opened 2026-06-30
adjust error message

amirafzali:
Summary: changing error message to smooth over some c10d tests that check for this phrasing (ideally they should not be hardcoding the error message wording)

Differential Revision: D110244474


d4l3k (approved):
lgtm

PR #522, opened 2026-09-15
Replace FD_INVALID macro with kInvalidFd constant to avoid collision with kernel header.

google-yfyang:
Linux kernel headers starting in Linux 6.12 (as well as Android NDK r30+) define:
`#define FD_INVALID -10009`
in `<linux/fcntl.h>`.
Because `gloo/transport/tcp/pair.cc` defines `#define FD_INVALID (-1)`, building against Linux 6.12+ kernel headers with `-Werror -Wmacro-redefined` fails with:
`error: 'FD_INVALID' macro redefined [-Werror,-Wmacro-redefined]`
Replace the preprocessor macro `FD_INVALID` with a `constexpr int kInvalidFd = -1;` constant inside the anonymous namespace in `pair.cc` to avoid polluting the macro namespace and colliding with kernel headers.

Test plan:

- `uvx --python 3.13 --managed-python clang-format==21.1.2 --dry-run --Werror gloo/transport/tcp/pair.cc`
- `git diff --check`


d4l3k:
@claude fix lint

d4l3k (approved):
LGTM

PR #523, opened 2026-09-16
Fix ThreadSanitizer data race in gloo::logLevel() initialization

jbspooner:
## Summary

Concurrent calls to `gloo::logLevel()` during startup race on the static `log_level` variable introduced in #467. Previously, `log_level` was initialized to `LogLevel::UNSET` in its static declaration and then read and assigned outside the C++11 static initialization guard (`__cxa_guard_acquire`), causing ThreadSanitizer data race warnings when multiple threads call `gloo::logLevel()` concurrently on startup.

Initialize `log_level` directly using an immediately-invoked lambda expression so that C++11 magic static initialization guarantees thread-safe, once-only evaluation.


d4l3k (approved):
LGTM

PR #525, opened 2026-09-18
fix(tcp, uv): call ntohs() when formatting port in Address::str()

kyx1999:
## Summary

Fix wrong port numbers in log and error messages produced by `Address::str()` in both the `tcp` and `uv` transports. The port was read directly out of the `sockaddr_in`/`sockaddr_in6` structure without converting it back to host byte order, so on little-endian machines every printed port appears with its two bytes swapped (e.g. port `384` is printed as `32769`, port `354` as `25089`).

## Motivation

We discovered this while debugging a connectivity failure: a distributed initialization built on gloo was failing with `ECONNREFUSED`, and the gloo error messages reported peer addresses whose ports did not match any port the application had ever configured. Because the printed port was byte-swapped, we initially misattributed the problem and suspected address corruption or misconfiguration in the initialization address book, and spent more time than we should have chasing that false lead before realizing the ports in the logs were simply printed in the wrong byte order — the underlying sockets and connection logic were always fine.

Since `Address::str()` is used pervasively in error paths (connect failures, timeouts, listener messages), this misleading output will send any developer who hits a transport-level failure down the same wrong path. Printing the correct port costs nothing and removes a real source of confusion when diagnosing connection problems.

## Root Cause

`sin_port`/`sin6_port` are stored in network byte order. `Address::str()` formatted them directly:

```cpp
// gloo/transport/tcp/address.cc (and the identical code in uv/address.cc)
port = in->sin_port;     // network byte order read as host order
port = in6->sin6_port;
```

On a little-endian host this swaps the two bytes of the port in every formatted string. For example, connecting to a peer listening on TCP port 384 prints `22.50.152.6:32769` (384 == 0x0180, stored in network byte order as 0x8001 and read raw on a little-endian host == 32769), which matches no port anything was configured to use.

## Fix

Call `ntohs()` when reading the port for formatting, in both transports:

- `gloo/transport/tcp/address.cc`: `Address::str()` for AF_INET and AF_INET6
- `gloo/transport/uv/address.cc`: `Address::str()` for AF_INET and AF_INET6, plus a `#include <arpa/inet.h>` for the `ntohs()` declaration

Only the display path is changed. Semantic uses of the port elsewhere (e.g. `htons()` when constructing addresses, and same-order comparisons such as the port subtraction in `Pair::isInitiator`) were audited and are correct as-is.

## Test Plan

- Built gloo with the patch; the tcp and uv transports compile cleanly.
- Ran a gloo-based multi-process initialization against an unreachable peer and confirmed connect-failure messages now print the real peer port (e.g. `...:5121` instead of the byte-swapped value).
- Existing test suite passes: `make test` (no behavior change beyond formatted output).

Branch: `kyx1999/fix-tcp-uv-address-str-ntohs`


kyx1999:
@d4l3k Could you take a look at this? Thanks!

d4l3k (approved):
Lgtm, good catch!

